PR #25, opened 2023-03-05
feat: add Dockerfile and docker-compose files

PabloCastellano:

PR #26, opened 2023-03-05
ci: publish docker image

PabloCastellano:
Before merging we should create a docker hub account and setup the credentials


0xGabi (approved):
We need to remember to merge #25 before this one.

Note that I already configured both `DOCKERHUB_USER` and `DOCKERHUB_TOKEN` as global organization secrets.

PR #41, opened 2023-04-05
feat: implement tRPC server and hook web client

0xGabi:
Build on #39


0xGabi:
Replaced by migration to monorepo

PR #67, opened 2023-05-06
Docker compose setup

PabloCastellano:
Run indexer, web and api in different containers.


Dockerfile
```diff
@@ -1,21 +1,40 @@
-FROM node:18.15-alpine as builder
+#FROM node:18-alpine as builder
```
0xGabi, line 1: I wonder if we missed to remove the commented sections of this file?

0xGabi (approved):
Well done!

PR #72, opened 2023-05-21
Added package-lock.json files

PabloCastellano:


PJColombo:
Oh my! Totally forgot about the lock file. Let's merge this

PJColombo:
My mistake. Doing a rollback of these changes.

I completely forgot that there's no need to add `package-lock.json` files as we're using `pnpm` which keeps a single lock file with all packages deps at root level.

PR #85, opened 2023-06-21
Tweaks and refactors

PJColombo:
It adds the following:

- Create a  blob storage manager component that orchestrates the instantiation and interactions with the different storage solutions (google storage and swarm)
- Make swarm set up optional
- Cron job schedule fix
- Share `days` library with a specific config across all packages/apps
- Fix `index` procedure issues related to the overall stats update logic
- Handle empty states on the React components
- Refactor and standardise tRPC query result handling and formatting
- Improve  packages/apps env variables parsing by using `zod`


packages/db/prisma/seed.ts
```diff
@@ -27,10 +27,10 @@ const UNIQUE_ADDRESSES_AMOUNT = 1000;
 const MAX_BLOB_BYTES_SIZE = 2048; // in bytes
 
 function buildGoogleStorageUri(hash: string): string {
```
0xGabi, line 29: I wonder if we should expose this from the storage manager as well?

packages/db/prisma/seed.ts
```diff
@@ -199,19 +202,9 @@ async function main() {
...
-      .exists();
-
-    if (!blobExists) {
-      await storage
-        .bucket(BUCKET_NAME)
-        .file(buildGoogleStorageUri(b.versionedHash))
-        .save(b.data);
-    }
-  });
+  const uploadBlobsPromise = uniqueBlobs.map(async (b) =>
```
0xGabi, line 205: I was not using the `UniqueBlobs` here to make sure we only store on swarm the blobs that we are actually storing on the database. If I'm not mistaken might be cases were some `UniqueBlobs` will not be used. Anyways is an edge case, we can go ahead with the current script.

0xGabi (approved):
Really awesome work! I learned a lot from reading the abstraction for the `BlobStorageManager`. You provide an excellent solution to the problem, and is future proof for when we start adding new storage services.

Also using `zod` for the handling env variables simplify a lot the logic, much clear now.

PR #102, opened 2023-07-17
Refactoring + Fixes

### LINKED ISSUE #96: Average blob size is calculated incorrectly
**Avg. Blob Size** can not be higher than **Total Blob Size**. One of them is incorrect.

image

PJColombo:
This PR contains the following main changes:

- Fix unique-elements overall metrics (`totalUniqueBlobs`, `totalUniqueSenders`, etc) calculation.
- Use [Prisma client extensions](https://www.prisma.io/docs/concepts/components/prisma-client/client-extensions) to encapsulate and expose stats aggregation and indexing logic.
- Create Vercel cron job endpoint for overall stats incremental updates.
- Prisma schema changes:
  - New first block fields to keep track of the block where an element appeared for the first time.
  - New `upsertedAt` and `insertedAt` fields
  - Remove redundant `id` fields
  - Create indexes for fk fields
- Incorrect blob metrics (total blob size and average blob size) display fixes (resolves #96) 
- ESLint changes:
  - Remove global config file
  - Create config files on `/packages/config/eslint` for different frameworks
  - Improve import sorting rules


0xGabi (approved):
Wow! Awesome work. I still need to give other pass to the stats extensions and the seed script that has the bulk of the changes. But looking good, let's go ahead and include the changes.

Issue #100, opened 2023-07-07
BEE_DEBUG_ENDPOINT should be optional


PabloCastellano:
It is used by the indexer but the web app does not use it.

Issue #98, opened 2023-07-07
Fail fast if specified blob storage manager credentials are incorrect


PabloCastellano:
* GCS => Try reading from the specified bucket using the given credentials
* Swarm => Try connecting to the endpoint and see if it is reachable (related: https://github.com/Blobscan/blobscan/issues/101)

PR #105, opened 2023-07-21
feat: agnostic blob storage manager

### LINKED ISSUE #97: Allow running blobscan without Google Cloud blob storage manager
* Allow blank string for all GOOGLE_* env vars
* Define sane defaults

```
blobscan-prod-indexer-1  | {"v":0,"name":"blobscan_indexer","msg":"Unexpected response from server","level":40,"hostname":"blobscan-vm","pid":1,"time":"2023-07-07T17:06:03.785908808Z","target":"blob_indexer::clients::blobscan","line":58,"
file":"src/clients/blobscan/mod.rs","url":"http://localhost:3002/api/indexer/block-txs-blobs","method":"PUT","body":"IndexRequest { block: Block { number: 30751, hash: 0x136d07e5560314a4316f3a756835e3e515f2ffb146e2d6a3b0b04f0b7470d782, ti
mestamp: 1688496120, slot: 30805 }, transactions: [Transaction { hash: 0x5a47c90bb4f364105239e9ec1b6094a72d0958c62c4c350402cdfa399fb6c15f, from: 0xd3248ba3e5492d767f8e427cb9c7b9d5c3972d7b, to: Some(0x000000000000000000000000000000000000f1
c1), block_number: 30751 }], blobs: [Blob { versioned_hash: 0x0175…8b2c, commitment: 0x978465e93f7e6ff37944bb83eb05f01589897758362ca13f64dfad8badfb7a7ccc8dc428b36402a4e8d092f4970aded1, tx_hash: 0x5a47…c15f, index: 0, data: [omitted] }] }"
,"response":"{\"message\":\"Failed to upload blobs to storages: A bucket name is needed to use Cloud Storage.\",\"code\":\"INTERNAL_SERVER_ERROR\"}"}
blobscan-prod-indexer-1  | {"v":0,"name":"blobscan_indexer","msg":"Failed to process slot. Retrying in 56 seconds…","level":40,"hostname":"blobscan-vm","pid":1,"time":"2023-07-07T17:06:03.785996903Z","target":"blob_indexer::slots_processo
r::slot_processor","line":38,"file":"src/slots_processor/slot_processor/mod.rs","slot":30805,"error":"ClientError(ApiError(ErrorResponse { code: String(\"INTERNAL_SERVER_ERROR\"), message: Some(\"Failed to upload blobs to storages: A bucket name is needed to use Cloud Storage.\") }))","target":"slot_processor"}
```

PJColombo:
Resolves #97, #98 and #100


packages/blob-storage-manager/src/index.ts
```diff
@@ -1,27 +1,40 @@
...
+const swarmStorage =
+  env.BEE_DEBUG_ENDPOINT && env.BEE_ENDPOINT
+    ? new SwarmStorage({
+        beeDebugEndpoint: env.BEE_DEBUG_ENDPOINT,
+        beeEndpoint: env.BEE_ENDPOINT,
+      })
+    : null;
+// Fallback to prisma if no storage is available
+const prismaStorage =
+  !googleStorage && !swarmStorage ? new PrismaStorage(prisma) : null;
```
PabloCastellano, line 25: We should give enough flexibility to configure the storages in any way, that is, instead of defining prismaStorage this way, I recommend using another environment variable so we can store to postgres and gcs.

PabloCastellano: Something like this would be ideal:

```
STORAGE_GCS_ENABLED=true
STORAGE_SWARM_ENABLED=false
STORAGE_POSTGRES_ENABLED=true
```

Then the storages would only be initialized if they are set to true. However if none is enabled we can keep this part so we fallback to postgres

.env.example
```diff
@@ -18,9 +18,14 @@ BEE_ENDPOINT=
 CHAIN_ID=7011893055
 
 GOOGLE_STORAGE_BUCKET_NAME=blobscan-test-bucket
-GOOGLE_STORAGE_PROJECT_ID=blobscan-379218
+GOOGLE_STORAGE_PROJECT_ID=blobscan-test-project
+GOOGLE_STORAGE_API_ENDPOINT=http://localhost:4443
 GOOGLE_SERVICE_KEY=
 
+GOOGLE_STORAGE_ENABLED=false
+PRISMA_STORAGE_ENABLED=false
```
PabloCastellano, line 26: Honestly I don't like the name :sweat_smile:. We know what prisma is because we are developing blobscan but probably someone that just wants to run it locally or deploy it to a kubernetes cluster is not familiar with Node.js ORMs and would need to look for what Prisma is. Also, what if we migrate to a different ORM in the future? It would make sense to rename this variable, right?

So far we are using Prisma as an ORM for Postgres databases. It probably works with prisma+mysql with small tweaks but we haven't tested it. For these reasons, I am inclined to think that `POSTGRES_STORAGE_ENABLED` is a better naming and more robust choice.

Lastly, let's use this as the default storage enabled, otherwise the blobs would not be stored anywhere.

In summary: `POSTGRES_STORAGE_ENABLED=true`

packages/api/src/router/indexer.ts
```diff
@@ -134,48 +130,54 @@ export const indexerRouter = createTRPCRouter({
       }
 
       // 2. Prepare address, block, transaction and blob insertions
-      const blockData = {
+
+      // TODO: Create an upsert extension that set the `insertedAt` and the `updatedAt` field
```
PabloCastellano, line 134: I'd create an issue too

packages/blob-storage-manager/src/env.ts
```diff
@@ -1,30 +1,61 @@
...
+      return z.NEVER;
+    }
+
+    return arg_ === "true";
+  };
+}
+
+const envSchema = z.object({
+  BEE_DEBUG_ENDPOINT: z.string().url().optional(),
+  BEE_ENDPOINT: z.string().url().optional(),
```
PabloCastellano, line 21: can we check somewhere that either BEE_ENDPOINT or BEE_DEBUG_ENDPOINT is defined when `SWARM_STORAGE_ENABLED=false`?

packages/blob-storage-manager/src/env.ts
```diff
@@ -1,30 +1,61 @@
...
+    .string()
+    .min(1)
+    .default("7011893055")
+    .transform((value, ctx) => {
+      const chainId = parseInt(value, 10);
+
+      if (isNaN(chainId) || chainId <= 0) {
+        ctx.addIssue({
+          code: z.ZodIssueCode.custom,
+          message: "CHAIN_ID must be a number greater than 0",
```
PabloCastellano, line 32: Proposal: `"CHAIN_ID is invalid: ${chainId}"`

packages/db/prisma/migrations/20230721002048_add_blob_data_and_data_reference_models/migration.sql
```diff
@@ -0,0 +1,46 @@
+/*
```
PabloCastellano, line 1: At some point we will have to include also migration scripts for production. But so far with devnet is fine

docker-compose.yml
```diff
@@ -12,6 +12,7 @@ services:
       - 25556:5556
     environment:
       BEACON_NODE_ENDPOINT: http://localhost:5052
+      GOOGLE_STORAGE_API_ENDPOINT: http://localhost:4443
```
PabloCastellano, line 15: We should pass the new env vars too. Maybe not all of them for the development docker-compose but just these:

```
GOOGLE_STORAGE_ENABLED=true
PRISMA_STORAGE_ENABLED=true
```

PabloCastellano (commented):
LGTM! Just left some comments about the env vars

.env.example
PJColombo: I agree. Let's change it

docker-compose.yml
PJColombo: Oops, forgot to do this

packages/blob-storage-manager/src/env.ts
PJColombo: You mean when `SWARM_STORAGE_ENABLED=true`? That check is performed [here](https://github.com/Blobscan/blobscan/blob/48573d9f163375f59ace9b6d907710d3f779c50a/packages/blob-storage-manager/src/storages/SwarmStorage.ts#L70)

packages/api/src/router/indexer.ts
PJColombo: Sure

packages/blob-storage-manager/src/env.ts
PabloCastellano: Nope, let me rephrase.

BEE_ENDPOINT is required by the web app while BEE_DEBUG_ENDPOINT is required by the indexer.

We should warn the user in case `SWARM_STORAGE_ENABLED=true` but BEE_DEBUG_ENDPOINT and BEE_ENDPOINT are undefined.

docker-compose.staging.yml
```diff
@@ -18,6 +18,8 @@ services:
       GOOGLE_STORAGE_BUCKET_NAME: "${GOOGLE_STORAGE_BUCKET_NAME}"
       GOOGLE_STORAGE_PROJECT_ID: "${GOOGLE_STORAGE_PROJECT_ID}"
       GOOGLE_SERVICE_KEY: "${GOOGLE_SERVICE_KEY}"
+      GOOGLE_STORAGE_ENABLED: "${GOOGLE_STORAGE_ENABLED}"
+      SWARM_STORAGE_ENABLED: "${SWARM_STORAGE_ENABLED}"
```
PabloCastellano, line 22: and `POSTGRES_STORAGE_ENABLED: "${POSTGRES_STORAGE_ENABLED}"`

.env.example
```diff
@@ -18,9 +18,14 @@ BEE_ENDPOINT=
 CHAIN_ID=7011893055
 
 GOOGLE_STORAGE_BUCKET_NAME=blobscan-test-bucket
-GOOGLE_STORAGE_PROJECT_ID=blobscan-379218
+GOOGLE_STORAGE_PROJECT_ID=blobscan-test-project
+GOOGLE_STORAGE_API_ENDPOINT=http://localhost:4443
```
PabloCastellano, line 22: Let's leave this as `GOOGLE_STORAGE_API_ENDPOINT=` since this variable is only used in development and it is currently hardcoded in the docker-compose.yml file

packages/blob-storage-manager/src/env.ts
PJColombo: Got it. Let's require users to provide at least one of these endpoints to instantiate the storage and throw error in case any of the operations which requires the unset client is called.

Also, let's add a similar warning to the google storage as well

PR #110, opened 2023-07-25
Use $BLOBSCAN_PROC env var instead of $1 in docker-entrypoint.sh

PabloCastellano:


Mandog1120:
9bae5a75287bd0a3c1b0fd679f22c461776e8ac2

PR #118, opened 2023-07-31
Blob storage fixes

### LINKED ISSUE #116: Do not require bee endpoints when swarm storage is disabled
The api won't run the following configuration, which should be valid:

```
BEE_DEBUG_ENDPOINT=
BEE_ENDPOINT=
SWARM_STORAGE_ENABLED=false
```

error:

```
ZodError: [                                                                              
  {                                                                                                                                                                                                             
    "validation": "url",
    "code": "invalid_string",                                                            
    "message": "Invalid url",
    "path": [                                                                            
      "BEE_DEBUG_ENDPOINT"                                                               
    ]                        
  },                                                                                     
  {                                                                                      
    "validation": "url",     
    "code": "invalid_string",                                                                             
    "message": "Invalid url",                                                            
    "path": [                                                                            
      "BEE_ENDPOINT"
    ]                   
  }                                   
]                                     
```

### LINKED ISSUE #117: Error uploading blobs to postgres storage
```
Failed to upload blobs to storages: Failed to upload blob to any of the storages :

-POSTGRES: PrismaClientKnownRequestError:

Invalid `.create()` invocation in
/app/packages/blob-storage-manager/src/storages/PostgresStorage.ts:39:8

  36   blobData: string
  37 ): Promise<string> {
  38   return this.client.blobData
  n→ 39     .create(
  Unique constraint failed on the fields: (`id`)
```

image

PJColombo:
It includes the following:

- Resolves #117
- Resolves #116
- New way to parse env variables  through `createEnvSchema()`. It abstracts common checks related to environment variables using Zod's schema.
- Improve blob storage manager errors description

### DIFF
--- a/apps/web/src/env.mjs
+++ b/apps/web/src/env.mjs
@@ -1,14 +1,22 @@
 import { createEnv } from "@t3-oss/env-nextjs";
 import { z } from "zod";
 
+import { optionalStringSchema } from "@blobscan/zod";
+
 export const env = createEnv({
   /**
    * Specify your server-side environment variables schema here. This way you can ensure the app isn't
    * built with invalid env vars.
    */
   server: {
-    DATABASE_URL: z.string().url(),
-    BEACON_NODE_ENDPOINT: z.string().url().default("http://localhost:3500"),
+    DATABASE_URL: optionalStringSchema(
+      z.string().url(),
+      "postgresql://blobscan:s3cr3t@localhost:5432/blobscan_dev?schema=public"
+    ),
+    BEACON_NODE_ENDPOINT: optionalStringSchema(
+      z.string().url(),
+      "http://localhost:3500"
+    ),
     NODE_ENV: z.enum(["development", "test", "production"]),
   },
   /**
@@ -16,8 +24,14 @@ export const env = createEnv({
    * For them to be exposed to the client, prefix them with `NEXT_PUBLIC_`.
    */
   client: {
-    NEXT_PUBLIC_EXPLORER_BASE_URL: z.string().url(),
-    NEXT_PUBLIC_BEACON_BASE_URL: z.string().url(),
+    NEXT_PUBLIC_EXPLORER_BASE_URL: optionalStringSchema(
+      z.string().url(),
+      "https://explorer.4844-devnet-7.ethpandaops.io/"
+    ),
+    NEXT_PUBLIC_BEACON_BASE_URL: optionalStringSchema(
+      z.string().url(),
+      "http://134.209.87.158:8080/"
+    ),
   },
   /**
    * Destructure all variables from `process.env` to make sure they aren't tree-shaken away.
--- a/apps/web/src/env.mjs
+++ b/apps/web/src/env.mjs
@@ -9,13 +9,10 @@ export const env = createEnv({
    * built with invalid env vars.
    */
   server: {
-    DATABASE_URL: optionalStringSchema(
-      z.string().url(),
-      "postgresql://blobscan:s3cr3t@localhost:5432/blobscan_dev?schema=public"
-    ),
+    DATABASE_URL: z.string().url(),
     BEACON_NODE_ENDPOINT: optionalStringSchema(
       z.string().url(),
-      "http://localhost:3500"
+      "http://localhost:5052"
     ),
     NODE_ENV: z.enum(["development", "test", "production"]),
   },
@@ -24,14 +21,8 @@ export const env = createEnv({
    * For them to be exposed to the client, prefix them with `NEXT_PUBLIC_`.
    */
   client: {
-    NEXT_PUBLIC_EXPLORER_BASE_URL: optionalStringSchema(
-      z.string().url(),
-      "https://explorer.4844-devnet-7.ethpandaops.io/"
-    ),
-    NEXT_PUBLIC_BEACON_BASE_URL: optionalStringSchema(
-      z.string().url(),
-      "http://134.209.87.158:8080/"
-    ),
+    NEXT_PUBLIC_EXPLORER_BASE_URL: z.string().url(),
+    NEXT_PUBLIC_BEACON_BASE_URL: z.string().url(),
   },
   /**
    * Destructure all variables from `process.env` to make sure they aren't tree-shaken away.
--- a/packages/api/src/env.ts
+++ b/packages/api/src/env.ts
@@ -1,7 +1,11 @@
-import { z } from "zod";
+import type { z } from "zod";
 
-export const env = z
-  .object({
-    SECRET_KEY: z.string().default("supersecret"),
-  })
-  .parse(process.env);
+import { createEnvSchema } from "@blobscan/zod";
+
+const envSchema = createEnvSchema({
+  SECRET_KEY: { default: "supersecret" },
+});
+
+export const env = envSchema.parse(process.env);
+
+export type Environment = z.infer<typeof envSchema>;
--- a/packages/api/src/env.ts
+++ b/packages/api/src/env.ts
@@ -1,9 +1,9 @@
-import type { z } from "zod";
+import { z } from "zod";
 
 import { createEnvSchema } from "@blobscan/zod";
 
 const envSchema = createEnvSchema({
-  SECRET_KEY: { default: "supersecret" },
+  SECRET_KEY: { schema: z.string() },
 });
 
 export const env = envSchema.parse(process.env);
--- a/packages/blob-storage-manager/src/BlobStorageManager.ts
+++ b/packages/blob-storage-manager/src/BlobStorageManager.ts
@@ -59,16 +59,34 @@ export class BlobStorageManager<
     const availableReferences = blobReferences.filter(
       ({ storage }) => this.#blobStorages[storage]
     );
-    return Promise.any(
-      availableReferences.map(({ reference, storage: storageName }) =>
-        (this.#blobStorages[storageName] as BlobStorage)
-          .getBlob(reference)
-          .then((data) => ({
-            data,
-            storage: storageName,
-          }))
-      )
-    );
+
+    try {
+      const blob = await Promise.any(
+        availableReferences.map(({ reference, storage: storageName }) =>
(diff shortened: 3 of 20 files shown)


packages/db/prisma/extensions/base.ts
```diff
@@ -101,19 +101,32 @@ export const baseExtension = Prisma.defineExtension((prisma) =>
...
-          return blobs.filter(
-            (b) => !existingBlobVersionedHashes.includes(b.versionedHash)
-          );
+          const newBlobVersionedHashes = Array.from(
+            new Set(blobs.map((b) => b.versionedHash))
+          ).filter((hash) => !dbBlobVersionedHashes.includes(hash));
+
+          return newBlobVersionedHashes.map((versionedHash) => {
+            // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+            const b = blobs.find((b) => b.versionedHash === versionedHash)!;
```
0xGabi, line 121: I think we are repeating the same logic twice here. Instead of having a `newBlobVersionedHashes` we can keep all the `newBlobs` in the first filter and return that. What do you think?

packages/zod/src/schemas.ts
```diff
@@ -0,0 +1,107 @@
...
+    if (arg_ !== "true" && arg_ !== "false") {
+      ctx.addIssue({
+        code: z.ZodIssueCode.custom,
+        message: `${varName} must be either true or false`,
+      });
+
+      return z.NEVER;
+    }
+
+    return arg_ === "true";
```
0xGabi, line 33: If we want to prevent the blob storages be set to `true` if the ENV were kept empty. Should not the default value be `false` here?

0xGabi (commented):
Looking good, left some minor comments 🙌

packages/blob-storage-manager/src/BlobStorageManager.ts
```diff
@@ -54,22 +59,40 @@ export class BlobStorageManager<
...
+        )
+      );
+
+      return blob;
+    } catch (e) {
+      const errorMessage = "Failed to get blob from any of the storages";
+
+      if (e instanceof AggregateError) {
+        const storageErrors = e.errors.map((err, i) => {
+          const storageName = availableReferences[i]?.storage ?? "Unknown";
```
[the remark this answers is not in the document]
PJColombo: Yes, in a normal case, I would check the errors to ensure they match the storage from which they were thrown. However, when dealing with the [`Promise.any`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/any#return_value) method, this isn't required. We receive an`AggregateError` when all the promises fail, and they maintain the same order as when executed.

I'll add a comment above this section in the code to make it clearer.

packages/blob-storage-manager/src/env.ts
```diff
@@ -1,59 +1,24 @@
...
+  CHAIN_ID: { schema: chainIdSchema(), default: 7011893055 },
+  GOOGLE_STORAGE_BUCKET_NAME: { optional: true },
+  GOOGLE_STORAGE_PROJECT_ID: { optional: true },
+  GOOGLE_SERVICE_KEY: { optional: true },
+  GOOGLE_STORAGE_API_ENDPOINT: {
+    schema: z.string().url(),
+    optional: true,
+  },
+  GOOGLE_STORAGE_ENABLED: { schema: booleanSchema(), default: false },
+  POSTGRES_STORAGE_ENABLED: { schema: booleanSchema(), default: false },
```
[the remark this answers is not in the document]
PJColombo: Yes, this was part of the discussions I had with @PabloCastellano. We believed it would be better for the user to explicitly set the storages they want to use.

packages/db/prisma/extensions/base.ts
PJColombo: We pass the blob's versioned hashes to be indexed through a set to remove any possible duplicates as a block can contain the same blob multiple times. Then we filter out the blobs that are already stored in the db.

Next, we iterate over these filtered versioned hashes to build the full blob again.

We can use the exclamation mark operator without any issues, as the resulting filtered blobs are a subset of the original blobs received

PJColombo: Perhaps we could add a comment here as well, to provide some clarity on this logic.

packages/zod/src/schemas.ts
PJColombo: If we use `"false"` here, ,then we would end up with the opposite result instead.

For example: say we set one of the enabled env vars to "true" (`arg_="true"`) then this zod transform fn would return false as `"true" === "false"` is false

packages/blob-storage-manager/src/env.ts
```diff
@@ -1,59 +1,24 @@
...
+import {
+  z,
+  chainIdSchema,
+  booleanSchema,
+  createEnvSchema,
+} from "@blobscan/zod";
+
+const envSchema = createEnvSchema({
+  BEE_DEBUG_ENDPOINT: { schema: z.string().url(), optional: true },
+  BEE_ENDPOINT: { schema: z.string().url(), optional: true },
```
PabloCastellano, line 10: Remember to check that either BEE_ENDPOINT or BEE_DEBUG_ENDPOINT are defined when SWARM_STORAGE_ENABLED=true as we spoke

packages/blob-storage-manager/src/BlobStorageManager.ts
```diff
@@ -90,29 +113,27 @@ export class BlobStorageManager<
...
-          "\n"
-        )}`
-      );
-    }
-
-    if (failedUploads.length) {
-      console.warn(
-        `Couldn't upload blob to some of the storages :\n ${storageErrors.join(
+        `Failed to upload blob ${versionedHash} to any of the storages :\n${storageErrors.join(
           "\n"
```
PabloCastellano, line 129: In my experience when debugging issues and reviewing logs, this \n will make it less readable. Do you mind joining with commas? (", ")

PabloCastellano (commented):
Gj!

I have left some comments about concerns with the env vars and their default values

packages/blob-storage-manager/src/env.ts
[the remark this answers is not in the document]
PabloCastellano: > Yes, this was part of the discussions I had with @PabloCastellano. We believed it would be better for the user to explicitly set the storages they want to use.

But for postgres, true is a saner default.

packages/blob-storage-manager/src/BlobStorageManager.ts
```diff
@@ -90,29 +113,27 @@ export class BlobStorageManager<
     const failedUploads = results.filter(
       (res) => res.status === "rejected"
     ) as PromiseRejectedResult[];
-    const storageErrors = failedUploads.map((res, i) => {
+    const storageErrors = failedUploads.map<StorageError<SNames>>((res, i) => {
       const storage = availableStorages[i] as [SNames, BlobStorage];
```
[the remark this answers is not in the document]
PJColombo: Oops, here you're definitely right.

I'll fix this

apps/web/src/env.mjs
```diff
@@ -1,23 +1,37 @@
...
     NODE_ENV: z.enum(["development", "test", "production"]),
   },
   /**
    * Specify your client-side environment variables schema here.
    * For them to be exposed to the client, prefix them with `NEXT_PUBLIC_`.
    */
   client: {
-    NEXT_PUBLIC_EXPLORER_BASE_URL: z.string().url(),
-    NEXT_PUBLIC_BEACON_BASE_URL: z.string().url(),
+    NEXT_PUBLIC_EXPLORER_BASE_URL: optionalStringSchema(
```
PabloCastellano, line 27: I would not make these variables optional neither. They are part of the minimum required for blobscan to work. We already provide default values in the .env.example file

packages/blob-storage-manager/src/env.ts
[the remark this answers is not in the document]
PJColombo: Oops, I got that backwards. I'll set this one to `true` then

packages/blob-storage-manager/src/env.ts
PJColombo: Yes, this was already [done](https://github.com/Blobscan/blobscan/blob/b2b2d10ac8ce6ed5104ffd4a257b319a7f2d90f6/packages/blob-storage-manager/src/storages/SwarmStorage.ts#L84) on previous PR as we discussed

packages/blob-storage-manager/src/BlobStorageManager.ts
PJColombo: Sure, no problem.

packages/blob-storage-manager/src/BlobStorageManager.ts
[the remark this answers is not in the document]
0xGabi: Nice, I did not took that into accont.

packages/db/prisma/extensions/base.ts
0xGabi: Cool, thank you for explain it. I did not factored the fact that we need the `Set` includes method to make filter easier.

packages/zod/src/schemas.ts
0xGabi: Right, I miss read it.

packages/api/src/env.ts
```diff
@@ -1,7 +1,13 @@
...
-  .parse(process.env);
+import { createEnvSchema } from "@blobscan/zod";
+
+const envSchema = createEnvSchema({
+  SECRET_KEY: { schema: z.string() },
+});
+
+export const env = envSchema.parse(process.env);
+
+console.log(env);
```
0xGabi, line 11:
```suggestion
```

0xGabi (approved):

packages/blob-storage-manager/src/env.ts
PabloCastellano: Oops thanks! I forgot it.

I have opened another PR fixing the logic. https://github.com/Blobscan/blobscan/pull/125

PR #125, opened 2023-08-02
Logic retrieving either BEE_ENDPOINT or BEE_DEBUG_ENDPOINT

PabloCastellano:
One of both must be given when enabling swarm blob storage.


PabloCastellano:
Does not make sense anymore after discussion

PR #138, opened 2023-08-17
feat: store blob gas fields

PJColombo:
It's related to the following indexer's [PR](https://github.com/Blobscan/blobscan-indexer.rs/pull/45)

It stores new blob gas fields sent by the indexer.


packages/blob-storage-manager/src/index.ts
```diff
@@ -3,6 +3,7 @@ import { env } from "./env";
 import { GoogleStorage, PostgresStorage, SwarmStorage } from "./storages";
 
 let blobStorageManager: BlobStorageManager | undefined;
+console.log(env);
```
0xGabi, line 6:
```suggestion
```

packages/api/src/utils/ethereum.ts
```diff
@@ -0,0 +1,53 @@
+const MIN_DATA_GASPRICE = 1;
+const DATA_GASPRICE_UPDATE_FRACTION = 3_338_477;
```
0xGabi, line 2:
```suggestion
const MIN_BLOB_GASPRICE = 1;
const BLOB_GASPRICE_UPDATE_FRACTION = 3_338_477;
```

packages/api/src/utils/ethereum.ts
```diff
@@ -0,0 +1,53 @@
...
+export function calculateBlobSize(blob: string): number {
+  return blob.slice(2).length / 2;
+}
+
+export function calculateBlobGasPrice(excessDataGas: bigint): bigint {
+  return BigInt(
+    fakeExponential(
+      BigInt(MIN_DATA_GASPRICE),
+      excessDataGas,
+      BigInt(DATA_GASPRICE_UPDATE_FRACTION)
```
0xGabi, line 50:
```suggestion
      BigInt(MIN_BLOB_GASPRICE),
      excessDataGas,
      BigInt(BLOB_GASPRICE_UPDATE_FRACTION)
```

packages/api/src/utils/ethereum.ts
0xGabi: Following the recent name update: https://github.com/ethereum/EIPs/commit/58e82a0addec750bc3799b05f771cd8ab3c7fae1

packages/api/src/utils/ethereum.ts
```diff
@@ -0,0 +1,53 @@
+const MIN_DATA_GASPRICE = 1;
```
0xGabi, line 1: Well done 👏

packages/db/prisma/schema.prisma
```diff
@@ -69,12 +69,15 @@ model BlobDataStorageReference {
...
-  slot       Int      @unique
-  insertedAt DateTime @default(now())
-  updatedAt  DateTime @default(now())
+  hash                  String   @id
+  number                Int      @unique
+  timestamp             DateTime @unique
+  slot                  Int      @unique
+  blobGasUsed           BigInt
+  excessBlobGas         BigInt
+  blobAsCalldataGasUsed BigInt
```
0xGabi, line 78: I was a bit confused with having the same naming for this field for both the block and tx. I'm not against it, though I wonder if something like `blobAsCalldataSum` could be more explicit 🤔

0xGabi (changes requested):
I really appreciate how clean and well-organized your code is. It made the review process smooth and straightforward. Great job! 💪

Left a few suggestions, other than that is ready to go.

packages/db/prisma/schema.prisma
PJColombo: I'd prefer to keep consistency with the field `blobGasUsed` which also contains a sum metric 🤔

PJColombo: `blobAsCalldata` still doesn't convince me entirely so maybe we can find a better name for it

PR #154, opened 2023-08-23
Add unit test to blob-storage-manager, api and db packages

0xGabi:


packages/blob-storage-manager/src/storages/GoogleStorage.ts
```diff
@@ -5,21 +5,21 @@ import type { BlobStorageConfig } from "../BlobStorage";
...
-type GoogleCredentials = {
+export type GoogleCredentials = {
   client_email: string;
   private_key: string;
 };
 
 export class GoogleStorage extends BlobStorage {
-  #storageClient: Storage;
-  #bucketName: string;
+  _storageClient: Storage;
```
0xGabi, line 21: @PJColombo what do you think of make this protected instead to access them in the mock subclass?

packages/dayjs/index.ts
```diff
@@ -1,6 +1,8 @@
 import dayjs from "dayjs";
 import relativeTime from "dayjs/plugin/relativeTime";
+import utc from "dayjs/plugin/utc";
```
0xGabi, line 3: @PJColombo let me know what you think of this change. During a tests I found out that we were using the local time instead.

packages/blob-storage-manager/src/BlobStorageManager.ts
```diff
@@ -116,7 +116,7 @@ export class BlobStorageManager<
               return `${storageName} - ${err}`;
             });
 
-            throw new Error(`${errorMessage}: ${storageErrors.join(",")}`);
+            throw new Error(`${errorMessage}: ${storageErrors.join(", ")}`);
```
0xGabi, line 119: Small nitpick to display better error messages.

packages/db/prisma/client.ts
```diff
@@ -0,0 +1,22 @@
+import { PrismaClient } from "@prisma/client";
```
0xGabi, line 1: Moving this logic to his own file was necessary to follow the vitest convention to mock modules.

packages/blob-storage-manager/src/BlobStorageManager.test.ts
```diff
@@ -0,0 +1,163 @@
...
+  const swarmStorage = new SwarmStorage(SWARM_STORAGE_CONFIG);
+
+  const blobStorageManager = new BlobStorageManager(
+    {
+      POSTGRES: postgresStorage,
+      GOOGLE: googleStorage,
+      SWARM: swarmStorage,
+    },
+    CHAIN_ID
+  );
```
PJColombo, line 33: Let's instantiate all of this in a `beforeAll` hook

packages/dayjs/index.ts
PJColombo: Cool. Nice catch 👍 !

packages/blob-storage-manager/src/index.test.ts
```diff
@@ -0,0 +1,62 @@
...
+    SwarmStorage: {
+      tryCreateFromEnv: vi.fn().mockResolvedValue(storages.SwarmStorage),
+    },
+    PostgresStorage: {
+      tryCreateFromEnv: vi.fn().mockResolvedValue(storages.PostgresStorage),
+    },
+  };
+});
+
+describe("createOrLoadBlobStorageManager", () => {
```
PJColombo, line 37: Let's keep this function untested for now.

These unit tests are too implementation specific so I'd just remove them

packages/db/prisma/extensions/base.test.ts
```diff
@@ -0,0 +1,69 @@
+import { describe, expect, it } from "vitest";
+
+import { updateAddressData } from "./base";
+
+describe("updateAddressData", () => {
```
PJColombo, line 5: Same as before: Let's avoid unit testing this internal function as well.

We can check this logic indirectly when testing for the extension functions that use this underneath

packages/db/prisma/extensions/base.ts
PJColombo, line 1: Keep this internal

packages/db/prisma/index.ts
```diff
@@ -1,23 +1,3 @@
...
-      process.env.NODE_ENV === "development"
-        ? ["query", "error", "warn"]
-        : ["error"],
-  });
-
-export const prisma = prisma_.$extends(baseExtension).$extends(statsExtension);
-
-export type BlobscanPrismaClient = typeof prisma;
-
+export * from "./client";
```
PJColombo, line 1: Maybe we can do the same for the blob storage manager package. Have a client file that holds the singleton

packages/db/prisma/seed/utils.test.ts
PJColombo, line 1: Logic related to local dev only. Let's test external behaviour only

packages/db/prisma/utils/dates.test.ts
PJColombo, line 1: Internal logic. No need to test it explicitly

PJColombo (changes requested):
Hey @0xGabi

Nice job on this PR. I can tell a lot of effort was put into this 💪 .

Left a series of review comments. I think almost all of them revolve around the same concept misunderstandings.

There's a major guideline I like to follow: **Focus on behaviour and requirements**

Usually, I think tests should focus on testing what a piece of code should do instead of  how it does it. This give us flexibility and smoothes the dev experience as we can do refactoring and changes  more freely without worrying about the tests complaining constantly because they're tied to closely to the internal implementation. 

This is why I like to avoid testing private or internal functions explicitly too. There are some cases where the internal logic is too complex and maybe it requires its own suite of tests tho.

Overall, good job. We just need to solve this problem to push this forward

package.json
```diff
@@ -17,6 +17,8 @@
     "format": "prettier --write \"**/*.{js,cjs,mjs,ts,tsx,md,json}\" --ignore-path .gitignore",
     "lint": "turbo lint && manypkg check",
     "lint:fix": "turbo lint:fix && manypkg fix",
+    "test:unit": "turbo test:unit",
```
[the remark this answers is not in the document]
0xGabi: Yes I agree, I move the configuration to use vitest workspace and a single test command.

packages/blob-storage-manager/src/storages/GoogleStorage.test.ts
```diff
@@ -0,0 +1,125 @@
...
+      expect(storage).toBeDefined();
+      expect(storage.bucketName).toEqual(GOOGLE_STORAGE_CONFIG.bucketName);
+      expect(storage.storageClient.projectId).toEqual(
+        GOOGLE_STORAGE_CONFIG.projectId
+      );
+      expect(storage.storageClient.apiEndpoint).toEqual(
+        GOOGLE_STORAGE_CONFIG.apiEndpoint
+      );
+
+      const [serviceAccont] = await storage.storageClient.getServiceAccount();
```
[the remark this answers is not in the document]
0xGabi: mmm I can't find the type 🤔

0xGabi:
I'll go ahead and merge this changes and will incorporate them in the integration branch. I just applied the suggestions and made a single pipeline for tests.

PR #159, opened 2023-08-29
feat: blob gas stats

PJColombo:
It adds the following:

- Create the following blob gas related metrics:
  - Block:
    - Total Blob Gas Used
    - Total Blob As Calldata Gas Used: The amount of gas the blobs in the block would have cost if it were send as calldata
    - Total Blob Fee
    - Total Blob As Calldata Fee
    - Avg. Blob As Calldata Fee
    - Avg. Blob Gas Price
  - Transaction:
    - Avg. Max Blob Gas Fee
 - Add charts to showcase previous metrics  
 - Add the following changes to the homepage:
   - Display gas expenditure comparison chart (between blob gas and blob as calldata gas)
   - Update displayed metrics
 - Add a new gas expenditure field to block and transaction details pages
 - Add a toolbox to charts containing the following functionality:
   - Chart data view
   - Switch between bar and line chart view
   - Switch to cumulative view
   - Download chart png image
 - Refactor stats procedures and router


PabloCastellano (commented):
I don't have any comment about the code.

However regarding the homepage look and feel, I think this part would look better if the data were organized in a different way:

image

I also see that we are not showing gas units (wei, gwei...) in some places such as this chart. These numbers are already big and devnet-8 has been running only for 2 weeks now:

image

Lastly, I tend to think that this chart in particular would look better as a bar chart.

Other than that, good job!

PJColombo:
> (maybe leaving the two charts to the right and the boxes to the left as before?)

Yes, I tried that layout but didn't feel quite right 🤔

Screenshot 2023-09-01 at 17 53 57

The layout seems a bit not well distributed 🤔

I think we can improve the look by making the charts a bit less detailed, similar to the one in etherscan's homepage 

image

PJColombo:
> I also see that we are not showing gas units (wei, gwei...) in some places such as this chart. 

Gas is unitless

> These numbers are already big and devnet-8 has been running only for 2 weeks now

Yeah, numbers are high but they add up. A new block containing 3 blobs is being validated every minute, that is: `~1440 daily blocks * 3 *  131072  blob gas  =  566231040 `

PR #172, opened 2023-09-05
Print environment variables on startup

PabloCastellano:


apps/rest-api-server/api/env.ts
```diff
@@ -6,7 +7,14 @@ export const env = createEnv({
     NODE_ENV: z.enum(["development", "test", "production"]).optional(),
     METRICS_ENABLED: booleanSchema.default("true"),
     TRACES_ENABLED: booleanSchema.default("false"),
+    PRINT_CONFIG: booleanSchema.default("false"),
```
PabloCastellano, line 10: Since this is only shown once on startup I would remove this option

apps/rest-api-server/api/instrumentation.ts
```diff
@@ -1,15 +1,11 @@
...
 
 import { setUpOpenTelemetry } from "@blobscan/open-telemetry";
 
 import { env } from "./env";
 
 if (env.TRACES_ENABLED) {
   setUpOpenTelemetry("blobscan_rest_api", {
-    instrumentations: [
-      new ExpressInstrumentation(),
-      new WinstonInstrumentation(),
```
PabloCastellano, line 12: Why did you remove WinstonInstrumentation? Is it related to this PR?

packages/blob-storage-manager/src/env.ts
```diff
@@ -12,9 +13,28 @@ export const env = createEnv({
     GOOGLE_STORAGE_ENABLED: booleanSchema.default("false"),
     POSTGRES_STORAGE_ENABLED: booleanSchema.default("true"),
     SWARM_STORAGE_ENABLED: booleanSchema.default("false"),
+    PRINT_CONFIG: booleanSchema.default("false"),
```
PabloCastellano, line 16: Same, I don't see the point of having it

PJColombo: Yeah you're right. I'll rollback this.

PR #173, opened 2023-09-05
Update BlobTransactionCard

0xGabi:
On mobile I notice the color is not right for the from value of the `BlobTransactionCard`


PJColombo (approved):
👍

Issue #168, opened 2023-09-04
Add feedback widget to web app


0xGabi:
An open source solution like: https://feedbackfin.com/setup.html could be interesting alternative to products like [freshdesk](https://www.freshworks.com/freshdesk/)

PR #182, opened 2023-09-27
feat(web): add feedback widget

0xGabi:
Solve #168 

Screenshot 2023-09-27 at 1 03 35 PM


apps/web/src/pages/api/feedback.ts
```diff
@@ -0,0 +1,52 @@
+/* eslint-disable no-case-declarations */
+import type { NextApiRequest, NextApiResponse } from "next";
+
+const webhooks = [
+  process.env.FEEDBACK_WEBHOOK_URL ??
+    "https://discord.com/api/webhooks/abc/123",
```
PabloCastellano, line 6: This is not a valid default so I'd rather remove it and make the variable optional

apps/web/src/pages/api/feedback.ts
```diff
@@ -0,0 +1,52 @@
...
+  const { message, rate, metadata } = req.body;
+  const method = req.method;
+
+  try {
+    switch (method) {
+      case "POST":
+        const text = `New Feedback ✨
+
+Message: ${message ? message : "-"}
+Rate: ${rate ? rateEmoji.get(rate) : "-"}
```
PabloCastellano, line 25: Should we make 'rate' mandatory?

PabloCastellano (commented):
I like the simplicity but have one concern, can we somehow rate-limit this endpoint to avoid spam?

turbo.json
```diff
@@ -51,6 +51,7 @@
     "CHAIN_ID",
     "CI",
     "DATABASE_URL",
+    "FEEDBACK_WEBHOOK_URL",
```
PabloCastellano, line 54: Documentation should be updated too

apps/web/src/pages/api/feedback.ts
0xGabi: This needs to be configured on the component. And there is no configuration for that allowed. To include this we should fork and make our own version of the widget.

PR #188, opened 2023-10-16
Add instrumentation for prometheus custom metrics

PabloCastellano:


packages/api/src/middlewares/withTelemetry.ts
```diff
@@ -1,43 +1,98 @@
-import type { api } from "@blobscan/open-telemetry";
+import { api } from "@blobscan/open-telemetry";
 
-import { tracer } from "../instrumentation";
+import { meter, tracer } from "../instrumentation";
 import { t } from "../trpc-client";
 
+type APICounter = {
```
PabloCastellano, line 6: I would merge APICounter with RequestsTotalCounter

packages/api/src/middlewares/withTelemetry.ts
```diff
@@ -1,43 +1,98 @@
...
+  "blobscan_api_requests_duration_ms",
+  {
+    description: "Duration of all the requests made to Blobscan API",
+    valueType: api.ValueType.INT,
+    unit: "ms",
+  }
+);
+
+function getEndpoint(url: string) {
+  return url.split("?")[0] ?? "";
```
PabloCastellano, line 35: Something like [this](https://stackoverflow.com/a/24006120) looks better to me

packages/api/src/middlewares/withTelemetry.ts
```diff
@@ -1,43 +1,98 @@
...
-    spanOptions.attributes = {
-      path: req.url ?? "unknown-path",
-    };
-  }
 
-  return tracer.startActiveSpan(spanName, spanOptions, async (span) => {
-    const result = await next();
+export const withTelemetry = t.middleware(
+  async ({ ctx: { req, res, scope }, next }) => {
+    const endpoint = req?.url ? getEndpoint(req.url) : "unknown-endpoint";
```
PabloCastellano, line 52: Is there any case in which cases req is not defined or the endpoint can't be determined from the url?

packages/api/src/middlewares/withTelemetry.ts
```diff
@@ -1,43 +1,98 @@
...
+      return result;
+    } finally {
+      const requestEnd = performance.now();
+
+      apiRequestsDurationMsHistogram.record(requestEnd - requestStart, {
+        scope,
+      });
+      apiRequestsTotalCounter.add(1, {
+        scope,
+        status_code: res?.statusCode?.toString() ?? "unknown-status-code",
```
PabloCastellano, line 92: Don't all responses have a status code?

packages/api/src/middlewares/withTelemetry.ts
```diff
@@ -1,43 +1,98 @@
...
+        endpoint,
+        method: req?.method ?? "unknown-method",
```
PabloCastellano, line 94: In which case does a http request not have method?

packages/blob-storage-manager/src/BlobStorageManager.ts
```diff
@@ -158,16 +169,25 @@ export class BlobStorageManager<
...
                     storage: name,
                   }));
+                const end = performance.now();
 
                 storageSpan.end();
 
-                return result;
+                updateBlobStorageMetrics({
+                  blobData: data,
+                  direction: "received",
```
PabloCastellano, line 185: should be "sent"

packages/blob-storage-manager/src/instrumentation.ts
```diff
@@ -1,3 +1,75 @@
...
+  meter.createHistogram<StorageMetricBaseAttributes>(
+    "blobscan_storage_transfer_duration_seconds",
+    {
+      description: "Duration of the transfers to/from a blob storage",
+      valueType: api.ValueType.INT,
+      unit: "ms",
+    }
+  );
+
+function calculateBlobSize(blob: string): number {
```
PabloCastellano, line 46: suggestion: calculateBlobBytes

apps/web/next.config.mjs
```diff
@@ -4,6 +4,14 @@ import "./src/env.mjs";
 /** @type {import("next").NextConfig} */
 const config = {
   reactStrictMode: true,
+  async rewrites() {
```
PabloCastellano, line 7: Could you add a comment about why this is needed?

packages/api/src/env.ts
```diff
@@ -1,10 +1,17 @@
...
+  nodeEnvSchema,
+  booleanSchema,
+} from "@blobscan/zod";
 
 export const env = createEnv({
   envOptions: {
     server: {
       SECRET_KEY: z.string(),
       NODE_ENV: nodeEnvSchema.optional(),
+      METRICS_ENABLED: booleanSchema.default("false"),
```
PabloCastellano, line 14: Current booleanSchema implementation already defaults to "false"

apps/web/src/env.mjs
```diff
@@ -1,6 +1,14 @@
 import { createEnv } from "@t3-oss/env-nextjs";
 import { z } from "zod";
 
+// See booleanSchema from packages/zod/src/schemas.ts
+// We need to redefine it here because we can't import ts files here
+const booleanSchema = z
+  .string()
+  .refine((s) => s === "true" || s === "false")
+  .transform((s) => s === "true")
+  .default("false");
```
PabloCastellano, line 10: Wdyt about removing the default here?

packages/api/src/middlewares/withTelemetry.ts
```diff
@@ -1,43 +1,98 @@
-import type { api } from "@blobscan/open-telemetry";
+import { api } from "@blobscan/open-telemetry";
 
-import { tracer } from "../instrumentation";
+import { meter, tracer } from "../instrumentation";
 import { t } from "../trpc-client";
 
+type APICounter = {
+  scope?: string;
```
PabloCastellano, line 7: Shouldn't scope be mandatory?

packages/api/src/middlewares/withTelemetry.ts
```diff
@@ -1,43 +1,98 @@
...
+const apiRequestsTotalCounter = meter.createCounter<RequestsTotalCounter>(
+  "blobscan_api_requests_total",
+  {
+    valueType: api.ValueType.INT,
+    description:
+      "Number of requests made to Blobscan API by endpoint, status_code and method",
+  }
+);
+
+const apiRequestsDurationMsHistogram = meter.createHistogram<APICounter>(
```
PabloCastellano, line 25: It should be `meter.createHistogram<RequestsTotalCounter>` too if I'm not wrong

packages/blob-storage-manager/src/BlobStorageManager.ts
```diff
@@ -81,18 +83,27 @@ export class BlobStorageManager<
...
                       storage: storageName,
                     }));
+                  const end = performance.now();
 
                   storageSpan.end();
 
-                  return result;
+                  updateBlobStorageMetrics({
+                    storage: storageName,
+                    blobData: blob.data,
```
PabloCastellano, line 101: suggestion:

```diff
-blobData: blob.data,
+blobSize: calculateBlobSize(blob.data),
```

PabloCastellano (commented):
:muscle:

packages/api/src/middlewares/withTelemetry.ts
PJColombo: Yes they do, but both responses and requests are instances of the same Node's `IncommingMessage` class which have this field set as optional as it is not defined on requests objects.

packages/api/src/middlewares/withTelemetry.ts
PJColombo: None. This was a workaround for running the tests on local dev  as they use a test context with no defined `req` and `res` objects.

I'll remove these unnecessary checks and fix the test context.

packages/api/src/middlewares/withTelemetry.ts
PJColombo: Yes. This happened for the same reason detailed above

packages/api/src/middlewares/withTelemetry.ts
PJColombo: There isn't a specific case. This check is in place to satisfy TypeScript typing.

In Node.js, both requests and responses are instances of the `IncomingMessage` class because they share many fields and methods. As a result, certain fields are marked as optional because they aren't applicable to both types (e.g. `method` on server responses).

PJColombo: I'll use an empty string instead of `unknown-status-code` to make it less confusing.

packages/api/src/middlewares/withTelemetry.ts
PJColombo: Sure

packages/blob-storage-manager/src/BlobStorageManager.ts
PJColombo: Oops

packages/api/src/env.ts
PJColombo: No, it doesn't.

It will throw an error if the env var is not defined

packages/api/src/middlewares/withTelemetry.ts
PJColombo: Yep

packages/api/src/middlewares/withTelemetry.ts
PJColombo: Yes, I think it's better

PR #191, opened 2023-10-27
refactor: refactor stats calculation scripts

PJColombo:


packages/db/test/extensions/stats-extension.test.ts
```diff
@@ -53,10 +55,26 @@ function runDailyStatsFunctionsTests(
         describe("when filling in daily stats", statsCalculationTestsSuite);
       }
 
-      it("should fill in stats for a multiple days period", async () => {
+      describe("when filling in stats for a multiple days period", () => {
+        it("should do it correctly when `DD-MM-YYYY` dates are passed", async () => {
```
PabloCastellano, line 59: Should it say YYYY-MM-DD?

packages/db/prisma/utils/dates.ts
```diff
@@ -8,25 +8,39 @@ export type RawDatePeriod = {
...
+    return { to: normalizeDailyDate(dayjs()) };
+
+  const normalizedDatePeriod = {
+    from: datePeriod.from && normalizeDailyDate(datePeriod.from, "startOf"),
+    to: datePeriod.to && normalizeDailyDate(datePeriod.to, "endOf"),
   };
+
+  const { from, to } = normalizedDatePeriod;
+
+  if (from && to && new Date(from) > new Date(to)) {
```
PabloCastellano, line 39: At this point, we are sure that both `from` and `to` are defined. Aren't we?

scripts/stats-aggregator/daily.ts
```diff
@@ -0,0 +1,143 @@
...
+  const result = await dailyStatsFn(operationParam).then((res) => {
+    if (operation_ === "deleteAll") {
+      return "All";
+    }
+
+    return typeof res === "number" ? res : res.count;
+  });
+
+  const formattedFrom = from
+    ? dayjs(from).format("DD-MM-YYYY")
```
PabloCastellano, line 93: I'd go for using YYYY-MM-DD as it is universal and can be compared using "<" and ">" operators

scripts/stats-aggregator/overall.ts
```diff
@@ -0,0 +1,246 @@
+import commandLineArgs from "command-line-args";
+import commandLineUsage from "command-line-usage";
+
+import { BlockNumberRange, prisma } from "@blobscan/db";
+
+import { deleteOptionDef, helpOptionDef } from "./common";
+
+const BEACON_NODE_ENDPOINT = process.env.BEACON_NODE_ENDPOINT;
+const DEFAULT_UNPROCESSED_BLOCKS_BATCH_SIZE = 1_000_000;
```
PabloCastellano, line 9: isn't it too big for being the default value?

packages/api/src/middlewares/withDate.ts
```diff
@@ -21,10 +20,12 @@ export const withDatePeriod = t.middleware(({ next, input }) => {
   const { from, to } = datePeriodSchema.parse(input) || {};
   const hasAtLeastOneDate = from || to;
 
-  const datePeriod: DatePeriod | undefined = hasAtLeastOneDate && {
-    from: from ? normalizeDate(from) : undefined,
-    to: normalizeDate(to ?? dayjs()),
-  };
+  const datePeriod =
```
PabloCastellano, line 23: If hasAtLeastOneDate is false, then datePeriod will be a boolean. Is it expected?

packages/db/test/extensions/stats-extension.test.ts
PJColombo: Yes

packages/db/prisma/utils/dates.ts
PJColombo: No. You can specify a date period without a starting or ending date so you need to check if they're defined

scripts/stats-aggregator/daily.ts
PJColombo: Yes. It should follow that format. I'll fix it

scripts/stats-aggregator/overall.ts
PJColombo: It's' an arbitrary value, still don't know what's an appropriate value for this operation.

Let's reduce it to 100k blocks

packages/api/src/middlewares/withDate.ts
PJColombo: If `datePeriod` is a boolean is gonna be `false` and then is gonna be treated as it's not defined, but to make it more clearer let's set it to `undefined` instead.

scripts/stats-aggregator/overall.ts
```diff
@@ -0,0 +1,246 @@
...
+}
+
+export async function incrementOverallStats({
+  targetBlockId,
+  batchSize = DEFAULT_UNPROCESSED_BLOCKS_BATCH_SIZE,
+}: {
+  targetBlockId: BlockId;
+  batchSize?: number;
+}) {
+  if (!!BEACON_NODE_ENDPOINT && targetBlockId === "finalized") {
```
0xGabi, line 120: The condition here is not doing what is supposed when `BEACON_NODE_ENDPOINT` is defined and the `targetBlockId === "finalized". We enter the condition anyways.

0xGabi (commented):
@PJColombo I found an issue with the overall script.

PR #201, opened 2023-11-12
feat: set up blob propagation mechanism 

PJColombo:
It builds on top of #191.

This PR introduces a mechanism for propagating blob data across various storage systems through `bullmq` sandboxed workers. The core of this functionality is encapsulated within a new entity called `BlobPropagator` which seamlessly handles the underlying logic.

The workflow for blob propagation is designed as a `bullmq` flow. Initially, blobs are temporarily stored on the disk. Subsequently, dedicated storage workers are tasked with reading these blobs and storing them across each designated storage service. Upon successful storage across all services, a new job is generated. This job is then picked up by a 'finaliser' worker. The finaliser's primary role is to remove the blobs from the disk and ensure that references to the blob storage are accurately recorded in the database.

This pull request introduces a convenient CLI tool designed to enhance interaction with BullMQ queues. The tool simplifies the processes of retrying and removing jobs, making these operations more accessible and user-friendly.


packages/blob-storage-manager/src/BlobStorageManager.ts
```diff
@@ -160,11 +171,12 @@ export class BlobStorageManager<
         },
       },
       async (span) => {
-        const availableStorages = Object.entries(this.#blobStorages).filter(
-          ([, storage]) => storage
-        ) as [SNames, BlobStorage][];
+        // If no storages are provided, use all the storages
```
PabloCastellano, line 174: Why? Usually you would expect that if no storages are provided, then none is used.

packages/blob-propagator/src/blob-propagator.ts
```diff
@@ -0,0 +1,34 @@
...
+
+import { createOrLoadBlobStorageManager } from "@blobscan/blob-storage-manager";
+import { BlobStorage } from "@blobscan/db";
+
+import { BlobPropagator } from "./BlobPropagator";
+import { env } from "./env";
+
+let blobPropagator: BlobPropagator | undefined;
+
+export async function createOrLoadBlobPropagator() {
```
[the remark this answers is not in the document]
PJColombo: Ok

packages/blob-propagator/src/blob-propagator.ts
```diff
@@ -0,0 +1,34 @@
...
+  if (!blobPropagator) {
+    const bsm = await createOrLoadBlobStorageManager();
```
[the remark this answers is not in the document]
PJColombo: Sure

packages/blob-storage-manager/src/BlobStorageManager.ts
PJColombo: This refers to the selected storages provided by the user.

I added a new parameter to the `storeBlob` operation that allows the user to provide specific storages where they want to store the blobs in. If no storage is provided we default to the normal behaviour which is trying to store the blob in all the storages. 

Maybe it should be better to rename the variable to `selectedStorages` to convey better this functionality

.env.example
```diff
@@ -73,3 +80,4 @@ OTLP_AUTH_PASSWORD=
 
 OTEL_EXPORTER_OTLP_PROTOCOL=
 OTEL_EXPORTER_OTLP_ENDPOINT=
+OTEL_DIAG_ENABLED=
```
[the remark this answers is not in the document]
PJColombo: To enable the Otel diagnostic loggers which were being set by default

packages/blob-propagator/src/utils.ts
```diff
@@ -0,0 +1,85 @@
...
+
+export function buildBlobDataFilePath(versionedHash: string) {
+  return `${versionedHash}.txt`;
+}
+
+export async function createBlobDataFile({ data, versionedHash }: Blob) {
+  const blobfilePath = buildBlobDataFilePath(versionedHash);
+
+  try {
+    if (await checkBlobDataFileExists(blobfilePath)) {
```
0xGabi, line 25: Note we were calling the checkBlobDataFileExists with a path instead of the versionedHash.

packages/blob-propagator/test/helpers.ts
```diff
@@ -0,0 +1,109 @@
+import { afterEach, beforeEach, expect, it } from "vitest";
+import type { SuiteFactory } from "vitest";
+
+import { getBlobStorageManager } from "@blobscan/blob-storage-manager";
+import type { BlobReference } from "@blobscan/blob-storage-manager";
+import type { BlobStorage, BlobStorage as BlobStorageName } from "@blobscan/db";
+
+import type { BlobPropagationSandboxedJob } from "../src/types";
+import { createBlobDataFile, removeBlobDataFile } from "../src/utils";
+import gcsStorage from "../src/worker-processors/gcs";
```
0xGabi, line 10: Let's rename `gcsStorage` for `gcsWorker`.

packages/blob-propagator/test/gcs-worker.test.ts
```diff
@@ -0,0 +1,28 @@
...
+
+import { runStorageWorkerTestSuite } from "./helpers";
+
+describe(
+  "GCS Worker",
+  runStorageWorkerTestSuite("GOOGLE", {
+    runBeforeAllFns() {
+      beforeAll(() => {
+        vi.stubEnv("CHAIN_ID", "70118930558");
+        vi.stubEnv("GOOGLE_STORAGE_ENABLED", "false");
```
0xGabi, line 11:
```suggestion
        vi.stubEnv("GOOGLE_STORAGE_ENABLED", "true");
```

0xGabi (changes requested):
Great work on the PR! Impressive use of BullMQ for blob propagation. Minor tweaks needed, but overall, fantastic job! 💪

packages/blob-propagator/test/gcs-worker.test.ts
0xGabi: Having a closer look at the tests. I believe we are encountering the following issue: https://github.com/vitest-dev/vitest/issues/2752

packages/blob-propagator/src/utils.ts
PJColombo: Oops, I refactor this part of the code to build the path in the check method itself and forgot to remove it from other places

PJColombo:
> I wonder if you considered using [Vercel KV](https://vercel.com/docs/storage/vercel-kv) as the Redis storage solution.

I didn't think about that at all. I suppose we could use a Vercel KV connection for the BullMQ workers if we were running the API in Vercel.

PR #211, opened 2023-11-18
fix: add our own function to convert a hexadecimal string to utf-8

PabloCastellano:
ethers.js implementation raises an exception as soon as one of the characters cannot be utf-8 decoded

See https://blobscan.com/blob/0x01eb50cbcd04004b89d42cbaeed19b9728d7498ada756c675f4f658e4d0f847a

image


0xGabi (approved):
💪

PR #223, opened 2024-01-20
fix: dev script & enhance .env files

0xGabi:


PJColombo (changes requested):
Let's comment out swarm's env vars as well and move the `init_bucket.ts` script into the `api` package as it is the only place where we're using it. Other than that, everything looks good 👍

PR #239, opened 2024-02-12
Use REDIS_URI instead of REDIS_{HOST,PORT,HOST,USERNAME}

PabloCastellano:


0xGabi (approved):
Using the `REDIS_URI` makes more sense. I have reviewed the changes and addressed the TODOs left:
- Resolved the type issues we had.
- Masked the password in the URI.

clis/blob-propagation-jobs-cli/src/Context.ts
```diff
@@ -16,8 +17,9 @@ export class Context {
   #finalizerQueue: Queue<BlobPropagationJobData>;
   #propagatorFlowProducer: FlowProducer;
 
-  constructor(storages: $Enums.BlobStorage[], connection: ConnectionOptions) {
+  constructor(storages: $Enums.BlobStorage[], redisUri: string) {
     const uniqueStorageNames = [...new Set(storages)];
+    const connection = new IORedis(redisUri, { maxRetriesPerRequest: null });
```
0xGabi, line 22: Note we are setting `maxRetriesPerRequest` as `null`. To prevent this warning: https://github.com/nocodb/nocodb/issues/2452

PR #265, opened 2024-03-08
fix(blob-storage-manager): add custom errors + manager initialisation changes

PJColombo:
It does the following:

- Creates custom errors for storages and the storage manager.
- Fixes storage manager initialization by attempting to create enabled storages and printing a warning if an error occurs, instead of throwing an exception.


packages/blob-storage-manager/src/storages/GoogleStorage.ts
```diff
@@ -29,42 +28,48 @@ export class GoogleStorage extends BlobStorage {
...
+      }
 
-    this._bucketName = bucketName;
+      this._bucketName = bucketName;
 
-    storageOptions.projectId = projectId;
+      storageOptions.projectId = projectId;
 
-    this._storageClient = new Storage(storageOptions);
+      this._storageClient = new Storage(storageOptions);
```
0xGabi, line 50:
```suggestion
      storageOptions.projectId = projectId;

      this._storageClient = new Storage(storageOptions);

      this._bucketName = bucketName;
```

0xGabi (approved):

PR #282, opened 2024-03-12
feat(web): add StorageBadge component

### LINKED ISSUE #262: feat(web): create blob storage badges
It would be nice to have a badge component for each available storage, displaying each on the blob details view (there should be a new field for this)  and the blob card component if the blob has been propagated to it.

The badge should contain the storage icon and name; it should also have a color related to one of the main coolers of the storage's logo.

Available storages at the moment:

- [ ]  Postgres
- [ ]  GCS (Google Cloud Storage)
- [ ]  Eth Swarm

0xGabi:
Resolves #262 

Screenshot 2024-03-12 at 2 23 12 AM


PJColombo (commented):
Closing this one and opening a new one here #304  to avoid doing more merge commits

PR #284, opened 2024-03-12
feat(web): add back navigation from blob to transactions

### LINKED ISSUE #238: Add back navegation from Blobs to Transaction and Blocks

0xGabi:
Resolves #238 

- [x] Evaluate if still make sense to allow navigation directly from blob to block.

Screenshot 2024-03-12 at 4 26 23 AM


PJColombo (approved):
Great job!

I polished the transaction hash link and added support to navigating back to block as well

PR #296, opened 2024-03-13
fix(api): calculate blob gas as calldata for all the blobs in a tx

0xGabi:


packages/api/src/routers/indexer/indexData.utils.ts
```diff
@@ -73,16 +73,18 @@ export function createDBTransactions({
...
   return transactions.map<WithoutTimestampFields<Transaction>>(
     ({ from, gasPrice, hash, maxFeePerBlobGas, to }) => {
-      const txBlob: IndexDataInput["blobs"][0] | undefined = blobs.find(
-        (b) => b.txHash === hash
-      );
+      const txBlobs = blobs.filter((b) => b.txHash === hash);
 
-      if (!txBlob) {
+      if (txBlobs.length === 0) {
         throw new Error(`Blob for transaction ${hash} not found`);
```
PJColombo, line 79:
```suggestion
        throw new Error(`Blobs for transaction ${hash} not found`);
```

PJColombo: A small nitpick. Feel free to amend the current commit

PJColombo (approved):
LGTM

Left just a small nitpick.

PR #359, opened 2024-04-01
feat(blob-propagator): add several changes

PJColombo:


packages/blob-storage-manager/src/storages/SwarmStorage.ts
```diff
@@ -63,16 +65,16 @@ export class SwarmStorage extends BlobStorage {
     return file.data.text();
   }
 
-  protected async _storeBlob(
-    chainId: number,
-    versionedHash: string,
-    data: string
-  ) {
+  protected async _removeBlob(reference: string): Promise<void> {
+    await this._swarmClient.bee.unpin(reference);
```
PabloCastellano, line 69: This won't work as you mean. Content can't be removed from Swarm once you upload it. Unpinning only affects local storage of the bee node

packages/blob-storage-manager/src/BlobStorage.ts
```diff
@@ -48,13 +48,21 @@ export abstract class BlobStorage {
     }
   }
 
-  async storeBlob(
-    chainId: number,
-    hash: string,
-    data: string
-  ): Promise<string> {
+  async removeBlob(uri: string): Promise<void> {
```
PabloCastellano, line 51: What is the use case for removing a blob from a storage?

PabloCastellano (changes requested):
I need some context on what are the benefits of moving the chain id parameter from BlobStorageManager to BlobStorage?

packages/blob-storage-manager/src/BlobStorage.ts
PJColombo: @PabloCastellano  This PR proposes a change to the blob propagator to enable specifying any blob storage as the temporary storage, rather than limiting it to just the file system. Therefore, all blob storages should implement this method.

packages/blob-storage-manager/src/storages/SwarmStorage.ts
PJColombo: Noted. I'll leave it either way as it's better than doing anything I suppose

PR #370, opened 2024-04-11
feat: decode Starknet blobs

PJColombo:
It adds the following:

- Set up a blob decoder package to handle all rollup blob decoding logic
- Adds Starknet blob decoding logic
- Adds a new view on the blob details web page to visualize Starknet's decoded blobs

image


PabloCastellano (commented):
Good job! I think you should add this to the Features section of the documentation. Also some README file explaining where the .wasm file comes from would be nice.

PR #374, opened 2024-04-15
fix(web): homepage improvements

PJColombo:
It adds several ui homepage improvements:

- Reposition latest blocks section vertically
- Improve entity cards:
  - Add rollup icon
  - Add proof and blob storage data downloadable icon to blob card
  - Add blob gas fees to blob transaction card
  - Add blob gas price and biob gas used to block card
  - Fix some responsive issues
  - Fix skeletons
  - Set a fixed height for all cards
  - Add a sliding animation to incoming and outgoing latest entity cards
  - Use shortest unit for ether amounts automatically.
- Standardise card headers
- Improve chart cards:
  - Animate loading skeleton
  - Adjust light theme colours
  - Reposition title
- Improve metric cards:
  - Adjust text font
- Resolve Blobscan logo's layout shift issue
- Rearrange nav menu items
- Fix search bar icon size on Firefox and safari browsers
- Display app's version instead of latest commit

image

image


apps/web/src/pages/index.tsx
```diff
@@ -244,24 +233,147 @@ const Home: NextPage = () => {
...
+                      <div className={CARD_HEIGHT} key={b.versionedHash}>
+                        <BlobCard blob={b} transactions={[b.tx]} compact />
+                      </div>
+                    ),
+                  }))}
+                />
+              )}
+            </div>
+          </Card>
+          {/* <Card
```
0xGabi, line 310: What do you think of just remove this now?

0xGabi (approved):
Great improvement from the current home page. I can't live without it now that I see this new version 🙂

One nitpick, on mobile the popover on the `Gas Expenditure` chart overflow the margin:

Screenshot 2024-04-16 at 6 20 01 PM

apps/web/src/pages/index.tsx
PJColombo: Oops. Yes, I completely forgot about this.

apps/web/src/components/Badges/StorageBadge.tsx
```diff
@@ -21,31 +17,31 @@ type StorageConfig = {
...
     },
   },
   swarm: {
-    icon: <SwarmIcon />,
+    icon: <StorageIcon storage="swarm" />,
     style:
       "bg-orange-100 hover:bg-orange-200 text-orange-800 hover:text-orange-900 dark:bg-orange-900 dark:text-orange-300 dark:hover:bg-orange-800 dark:hover:text-orange-200",
     buildDownloadUrl(blobReference) {
-      return `https://gateway.ethswarm.org/access/${blobReference}`;
+      return buildStorageDownloadUrl("swarm", blobReference) ?? "#";
```
PabloCastellano, line 40: According to buildStorageDownloadUrl() implementation, when first parameter is "swarm" it will always return a string so I guess we can safely remove the `?? "#"` part

PabloCastellano (approved):
Great work! Just left a minor comment

PR #382, opened 2024-04-22
Some improvements in documentation

PabloCastellano:
- Removed cron jobs page (we don't use them anymore)
- Added background jobs page
- Expand CLI to "Command line interface"
- Updated list of grants received
- Mention bullmq as part of Blobscan architecture


apps/docs/src/app/docs/cron-jobs/page.md
```diff
@@ -1,31 +0,0 @@
----
```
0xGabi, line 1: We are deleting the file instead of rename it to `background-jobs`.

PabloCastellano: done

PR #415, opened 2024-06-10
fix: remove Average Blob Size metric

PabloCastellano:
This is not needed since blobs are always 128KB.

### Checklist


PJColombo (approved):
Overall, LGTM.

I made the following changes:

	•	Added the missing Prisma migration file
	•	Added the missing changeset
	•	Fixed an issue where the average max blob gas fee was removed from the transaction stats aggregation. This is something we want to keep.

remarcable:
@PabloCastellano Just a quick question – what did the metric show before it was removed?

So far I assumed it showed how "full" the blobs were, i.e. the % of the filled blob (`blob size` - `null bytes`). Is this correct?

So if you posted only 64KB of data to a blob and filled the rest with null bytes, then this blob would be 50% full. The metric here would aggregate this over all blobs.

Or did it show something else, e.g. the avg blob size in terms of "avg # of blobs per block * 128 KB`?

Thanks a lot!

PabloCastellano:
@remarcable I can't tell you exactly what this metric showed before it was removed. It was incorrect and didn't show the blob size - null bytes metrics you believed (however this is a very interesting metric that we can definitely add!).

We introduced this metric in the very beginning of the project when we didn't know the spec well enough and thought that the blob size wasn't fixed.

remarcable:
Thanks @PabloCastellano, this was insightful!

PR #422, opened 2024-07-01
feat: add block number and timestamp to transaction model

PJColombo:
### Checklist

### Description

It adds two main changes:

- It creates two new fields `block_number` and `block_timestamp` in the `transaction` table and populates them. It also creates an index for each of them. 
- It enables Prisma's relational join [strategy](https://www.prisma.io/blog/prisma-orm-now-lets-you-choose-the-best-join-strategy-preview).

This PR addresses the performance issues with the current Prisma query used to retrieve multiple transactions. The existing approach is inefficient because it retrieves full transaction data before filtering or sorting, leading to unnecessary overhead.


packages/db/prisma/client.ts
```diff
@@ -28,7 +28,7 @@ if (!prisma_) {
   // eslint-disable-next-line turbo/no-undeclared-env-vars
   if (process.env.MODE !== "test") {
     p.$on("query", (e) => {
-      logger.debug(`${e.query}\nDuration=${e.duration}ms`);
+      logger.info(`${e.query}\nDuration=${e.duration}ms`);
```
PabloCastellano, line 31: Remember to change it back to debug!

PR #423, opened 2024-07-04
feat: add block hash, number and timestamp field to `BlobsOnTransactions` model

PJColombo:
### Checklist

### Description

It builds on top of PR #422.

It creates two new fields `block_hash`, `block_number` and `block_timestamp` in the `blobs_on_transactions` table and populates them. It also creates an index for each of them.

This PR addresses the performance issues with the current Prisma query used to retrieve multiple blobs. The existing approach is inefficient because it retrieves full blob data before filtering or sorting, leading to unnecessary overhead.


packages/db/prisma/schema.prisma
```diff
@@ -3,7 +3,7 @@
 
 generator client {
   provider        = "prisma-client-js"
-  previewFeatures = ["tracing", "metrics"]
+  previewFeatures = ["relationJoins", "tracing", "metrics"]
```
PabloCastellano, line 6: Are we using `relationJoins` in the end?

packages/api/test/__snapshots__/blob.test.ts.snap
```diff
@@ -55,9 +55,6 @@ exports[`Blob router > getAll > when getting expanded blob results > should retu
     "index": 2,
     "proof": "proof004",
     "size": 1300,
-    "transaction": {
-      "rollup": "optimism",
-    },
```
PabloCastellano, line 60: why is this data being removed?

packages/api/test/__snapshots__/blob.test.ts.snap
```diff
@@ -36,11 +36,11 @@ exports[`Blob router > getAll > when getting expanded blob results > should retu
       "blobGasPrice": "22",
       "blobGasUsed": "5500000",
       "excessBlobGas": "15000",
-      "hash": "blockHash007",
       "slot": 107,
-      "timestamp": "2023-08-31T14:00:00.000Z",
     },
+    "blockHash": "blockHash007",
     "blockNumber": 1007,
+    "blockTimestamp": "2023-08-31T14:00:00.000Z",
```
PabloCastellano, line 43: Is this a breaking change for the API users? In this case, we should warn it

packages/db/prisma/schema.prisma
PJColombo: No, I accidentally left it in while testing. Good catch.

It looks is less performant compared to the current Prisma relation strategy when you try to pull elements from the db with a large offset.

packages/api/test/__snapshots__/blob.test.ts.snap
PJColombo: It shouldn't be a breaking change.

We're keeping the current structure and just adding a few extra fields

PR #456, opened 2024-07-29
feat(web): replace list of `BlobCard` by new `PaginatedTable` on the blobs view

### LINKED ISSUE #450: Use tables instead of cards in blobs page
Reuse [`Table`](https://github.com/Blobscan/blobscan/blob/next/apps/web/src/components/Table/index.tsx) component for this.

You can see an example of the table usage [here](https://staging.blobscan.com/blob/0x01e6065f6d70b5003eafc0aff42effc9b47aa3d375af892a3cfb2362904eb502).

Columns should be:

1. **Versioned Hash**
2. **Transaction Hash**: Link to the transaction
3. **Block Number**: Link to the block
4. **Timestamp**
5. **Size**
6. **Storage References**: Set of storage icons

xFJA:
### Checklist

### Description

It resolves #450 

- Create the new `PaginatedTable` component to list items in a table with pagination to be reused in `blob`, `block` and `transaction` views
- Migrate the current 'blob view' using a blob card list to the new `PaginatedTable` component

### Screenshots 

image


apps/web/src/components/PaginatedTable/helpers/blobs.tsx
PJColombo, line 1: I feel like this file has the following drawbacks:

1. This seems like premature refactoring as a helper function has been added that is only used on the`blobs` page view and is likely to be used only once.
2. This structure makes code discoverability difficult. We need to navigate to this file every time we want to review the blobs page view, and it is not immediately clear that this file contains the props.
3. We have the component implementation place in the same directory as a specific use case.


I would move this to the `blobs` page view

apps/web/src/components/PaginatedTable/helpers/blobs.tsx
```diff
@@ -0,0 +1,109 @@
...
+          versionedHash,
+          size,
+          dataStorageReferences,
+          txHash,
+          blockTimestamp,
+          blockNumber,
+        }) => ({
+          cells: [
+            {
+              item: shortenAddress(versionedHash, 8),
```
PJColombo, line 60: Let's turn this into a link to the blob

apps/web/src/pages/blobs.tsx
```diff
@@ -1,20 +1,26 @@
...
+import {
+  blobsTableHeaders,
+  getBlobsTableRows,
+} from "~/components/PaginatedTable/helpers";
 import { api } from "~/api-client";
 import { formatNumber } from "~/utils";
 
 const Blobs: NextPage = function () {
   const router = useRouter();
   const { p, ps } = getPaginationParams(router.query);
```
PJColombo, line 17: Now that we're using tables and we're able to show more elements using less space. I would modified the default page size to 50

PJColombo (changes requested):
Good work!

Left a few comments. There is some refactoring I feel we could do and some missing details we should have.

PJColombo:
Also, I feel the table header is shifting awkwardly during the transition of the skeleton and the actual data being rendered. Maybe we can add support to parametrize fixed widths to every column in the table?. This way we have a smoother transition 🤔 

https://github.com/user-attachments/assets/de8e5b07-62ea-4dda-b14e-0ba18a40821c

PJColombo:
One last thing: can we leave the pagination skeleton as it was before

That is, displaying a skeleton for the total pages only

image


We'll change later as we're moving to cursor-based pagination and we'll separate the total items endpoint call from the list of elements one. But we can leave it as it was for now

apps/web/src/components/PaginatedTable/helpers/blobs.tsx
```diff
@@ -0,0 +1,109 @@
...
+              item: (
+                <div className="text-contentTertiary-light dark:text-contentTertiary-dark">
+                  <Link href={buildBlockRoute(blockNumber)}>{blockNumber}</Link>
+                </div>
+              ),
+            },
+            {
+              item: (
+                <div className="min-w-60 whitespace-break-spaces">
+                  {formatTimestamp(blockTimestamp)}
```
PJColombo, line 81: Timestamp seems a little bit verbose. Can we show only the `x ago` part and omit the full date?

xFJA: Done! Added a `compact` param for the util

apps/web/src/pages/blobs.tsx
xFJA: Done :+1:

apps/web/src/components/PaginatedTable/helpers/blobs.tsx
xFJA: Done :+1:

apps/web/src/components/PaginatedTable/helpers/blobs.tsx
xFJA: Done :+1:

PJColombo (approved):

PR #462, opened 2024-07-30
refactor: Remove BigIntLike

luis-herasme:
### Checklist

### Description

This change addresses an error in the `calculatePercentage` function that occurred when using synthetic data generated by the data generator. The error message was:
```RangeError: The number 21460778973.687706 cannot be converted to a BigInt because it is not an integer```

To resolve this issue, I made the following changes:
1. Updated functions that previously accepted `BigIntLike` parameters to exclusively accept `bigint`.
2. Removed the `BigIntLike` type entirely from the codebase.
3. Added `numberToBigInt`

These modifications ensure type consistency and prevent potential conversion errors when working with large numbers.

This function takes two `BigIntLike`. Initially, I considered implementing a more robust conversion, which would have introduced some overhead. However, upon further review, I noticed that the function was only being used with `bigint`, except for one instance where we used a number, which is the same place where the original error occurred.

To address this, I updated the only two functions that took `BigIntLike` as a parameter to take `bigint` exclusively and removed the `BigIntLike` type altogether.

image

### DIFF
--- a/apps/web/src/utils/ethereum.ts
+++ b/apps/web/src/utils/ethereum.ts
@@ -1,4 +1,9 @@
-import { calculatePercentage, formatNumber, performDiv } from "./number";
+import {
+  calculatePercentage,
+  formatNumber,
+  numberToBigInt,
+  performDiv,
+} from "./number";
 
 export const GAS_PER_BLOB = 131_072; // 2 ** 17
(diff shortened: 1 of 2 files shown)


PJColombo (approved):
Nice job!

LGTM!

Issue #66, opened 2023-05-03
Create a custom page for http 404 and 500 errors


PJColombo:
[example](https://blobscan.com/search?q=0x8f90b7793272e0a59a8890197f6b1c1c14ba76fcbdfa139241a2ff4ddf504fce9c45c84fbcc3f0ce666c748159b57a39)

![image](https://github.com/user-attachments/assets/6829a987-13d6-48da-aa12-b0f26d0b7c54)

PabloCastellano:
Dunno if related but I found this error today in vercel logs, mentioning a 500.html not found file:

The link that caused this error was [a simple search](https://blobscan.com/search?q=0x8f90b7793272e0a59a8890197f6b1c1c14ba76fcbdfa139241a2ff4ddf504fce9c45c84fbcc3f0ce666c748159b57a39).

```
ching cause stack trace ...
    at createRetryer (file:///var/task/node_modules/@tanstack/query-core/build/lib/retryer.mjs:149:5) {
  code: 'INTERNAL_SERVER_ERROR',
  name: 'TRPCError',
  page: '/search',
  [cause]: TypeError: Cannot read properties of undefined (reading 'url')
      at buildURL (/var/task/apps/web/.next/server/pages/search.js:985:24)
      at /var/task/apps/web/.next/server/pages/search.js:992:17
      at callRecursive (file:///var/task/node_modules/@trpc/server/dist/index.mjs:456:38)
      at resolve (file:///var/task/node_modules/@trpc/server/dist/index.mjs:486:30)
      at callProcedure (file:///var/task/node_modules/@trpc/server/dist/config-00ffd309.mjs:181:12)
      at Object.query (file:///var/task/node_modules/@trpc/react-query/dist/server/index.mjs:20:28)
      at Object.queryFn (file:///var/task/node_modules/@trpc/react-query/dist/server/index.mjs:58:39)
      at Object.fetchFn [as fn] (file:///var/task/node_modules/@tanstack/query-core/build/lib/query.mjs:252:27)
      at run (file:///var/task/node_modules/@tanstack/query-core/build/lib/retryer.mjs:101:31)
      at createRetryer (file:///var/task/node_modules/@tanstack/query-core/build/lib/retryer.mjs:149:5)
}
MissingStaticPage [Error]: Failed to load static file for page: /500 ENOENT: no such file or directory, open '/var/task/apps/web/.next/server/pages/500.html'
    at /var/task/node_modules/next/dist/server/require.js:109:19
[...]

PR #463, opened 2024-07-31
fix: Fix search

luis-herasme:
### Checklist

### Description

The search bar currently only works if the suggestions are loaded. If the suggestions aren't loaded and you press the `enter` key, you'll be redirected to an error 500 page after a few seconds.

The problem occurred because an invalid `trpc context` was being created inside `getServerSideProps` in the `search.tsx` file.

As we are already on the server, there is no need to recreate a `trpc helper` each time someone visits `\search`. We can simply call the search function directly. To achieve this, I refactored the search functionality out of the `trpc query` so that we can use the search function both inside and outside the query.

### Related Issue (Optional)

https://github.com/Blobscan/blobscan/issues/66#issuecomment-2057009689

image
If you are on the main branch and you try to search for something, you will encounter this error. This error occurs inside the `withTelemetry` middleware. You might initially think it's unrelated to the search, but the issue actually stems from the middleware utilizing undefined data due to the `trpc context` being incorrectly defined inside the `getServerSideProps`.

### DIFF
--- a/apps/web/src/pages/search.tsx
+++ b/apps/web/src/pages/search.tsx
@@ -1,7 +1,7 @@
 import type { GetServerSideProps } from "next";
 import { useRouter } from "next/router";
 
-import { searchByTerm } from "@blobscan/api/src/routers/search/byTerm";
+import { searchByTerm } from "@blobscan/api";
 
 import { Button } from "~/components/Button";
 import type { RouterOutputs } from "~/api-client";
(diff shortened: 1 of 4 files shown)


apps/web/src/pages/search.tsx
```diff
@@ -1,10 +1,7 @@
 import type { GetServerSideProps } from "next";
 import { useRouter } from "next/router";
-import { createServerSideHelpers } from "@trpc/react-query/server";
-import superjson from "superjson";
 
-import type { TRPCContext } from "@blobscan/api";
-import { appRouter, createTRPCInnerContext } from "@blobscan/api";
+import { searchByTerm } from "@blobscan/api/src/routers/search/byTerm";
```
PJColombo, line 4: Let's add this function to the root `index.ts` in the api package

PJColombo (changes requested):
LGTM!

I left a tiny change request

Feel free to squash & merge this once we resolve that

Issue #471, opened 2024-08-02
Create "Next" and "Previous" buttons


PJColombo:
Create navigation buttons to go to previous and next element for the following entities:

- [ ] Block
- [ ] Transaction
- [ ] Blob

PR #479, opened 2024-08-03
feat: Add NavArrows component

### LINKED ISSUE #472: Create "Next" and "Previous" buttons for the block details page

luis-herasme:
### Checklist

### Description
Add NavArrows component

### Related Issue (Optional)
#471
Closes #472

image
image


.changeset/sweet-melons-beg.md
```diff
@@ -0,0 +1,5 @@
+---
+"@blobscan/web": minor
+---
+
+feat: Add NavArrows component
```
PJColombo, line 5: Changelog entries should not follow the commit format; they should be written in past tense.

Can we add something like "Added navigation arrow buttons to navigate to previous and next block"?

apps/web/src/components/NavArrow.tsx
```diff
@@ -0,0 +1,43 @@
...
+      <NavArrow type="Next" href={next} />
+    </div>
+  );
+}
+
+function NavArrow({ type, href }: { type: "Next" | "Prev"; href?: string }) {
+  const router = useRouter();
+
+  return (
+    <button
```
PJColombo, line 17: Can we use the button component we already have?

apps/web/src/pages/block/[id].tsx
```diff
@@ -77,7 +80,24 @@ const Block: NextPage = function () {
...
+            {blockData.number}
+            {latestBlock && (
+              <NavArrows
+                prev={buildBlockRoute(blockData.number - 1)}
+                next={
+                  latestBlock.number > blockData.number
+                    ? buildBlockRoute(blockData.number + 1)
+                    : undefined
+                }
+              />
```
Comment (PJColombo) [inline apps/web/src/pages/block/[id].tsx:96]: Blobscan only stores blocks that contain blob transactions, so the next block may not be the immediately preceding or following one

apps/web/src/pages/block/[id].tsx
```diff
@@ -47,6 +49,7 @@ const Block: NextPage = function () {
 
     return deserializeFullBlock(rawBlockData);
   }, [rawBlockData]);
+  const { data: latestBlock } = api.block.getLatestBlock.useQuery();
```
Comment (PJColombo) [inline apps/web/src/pages/block/[id].tsx:52]: We're already fetching this data in the `ExplorerDetails` component. 

If we need to access it in other parts of the component tree, it might be better to create a context now so we can consume it wherever necessary. What do you think?
wdyt?

PJColombo (changes requested):
Nice job!

It still needs a few changes.

.changeset/sweet-melons-beg.md
luis-herasme: Done d482a1e

apps/web/src/components/NavArrow.tsx
```diff
@@ -0,0 +1,43 @@
...
+export function NavArrows({ next, prev }: { next?: string; prev?: string }) {
+  return (
+    <div className="flex items-center justify-center gap-1">
+      <NavArrow type="Prev" href={prev} />
+      <NavArrow type="Next" href={next} />
+    </div>
+  );
+}
+
+function NavArrow({ type, href }: { type: "Next" | "Prev"; href?: string }) {
```
xFJA, line 13:
```suggestion
function NavArrow({ type, href }: { type: "next" | "prev"; href?: string }) {
```

apps/web/src/components/NavArrow.tsx
```diff
@@ -0,0 +1,43 @@
+import { useRouter } from "next/router";
```
xFJA, line 1: The filename should be called `NavArrows.tsx` as it's the component exported

apps/web/src/components/NavArrow.tsx
luis-herasme: Done [d894614](https://github.com/Blobscan/blobscan/pull/479/commits/d8946140f904f84c4a8fffadbb5e3c180e9accc0)

apps/web/src/components/NavArrow.tsx
luis-herasme: Done [7d5cc60](https://github.com/Blobscan/blobscan/pull/479/commits/7d5cc602581d4ba6d82569ee6de0bc8577f165fd)

apps/web/src/pages/block/[id].tsx
Comment (luis-herasme) [reply apps/web/src/pages/block/[id].tsx]: Done [83dcc29](https://github.com/Blobscan/blobscan/pull/479/commits/83dcc2943977815edb28f3e55070c47de385decf)

apps/web/src/pages/block/[id].tsx
Comment (luis-herasme) [reply apps/web/src/pages/block/[id].tsx]: We no longer use `latestBlock` after commit [83dcc29](https://github.com/Blobscan/blobscan/pull/479/commits/83dcc2943977815edb28f3e55070c47de385decf).

apps/web/src/components/NavArrow.tsx
luis-herasme: I'll make that change after the button refactorization PR is merged, since the style of this button is a bit different

xFJA (approved):

PJColombo (changes requested):
Great work.

However, I believe there might be a more efficient approach to retrieving neighboring blocks. Currently, we’re fetching the neighboring block numbers initially, only to fetch the entire block data again when clicking on the arrow buttons to navigate. This double-fetching could be optimized.

Additionally, rendering the arrow buttons after fetching the block by triggering the `getBlockNeighbours` procedure affects the UI experience, making it feel slightly disjointed. 

I think we can optimiser the overall logic by leveraging the tRPC React Query's cache , and fetching neighbour blocks on demand only 

We could update the `getBlockNeighbors` procedure by having it return the same block data structure as the `getBlockById` procedure. Additionally, we could introduce a new parameter to specify which neighboring block to retrieve (either previous or next) instead of retrieving both.

The flow I envision is as follows:

1.	The user clicks on either the “next” or “previous” arrow button.
2.	The user is redirected to a new page at `/block/[block-number]/neighbour?direction=[next|previous]` where the following server-side operations are performed:
  1.	Call `getBlockNeighbor` to retrieve the appropriate neighboring block (next or previous).
  2.	Update the React Query cache entries for the procedure calls `getBlockNeighbor` and `getBlockById` for the most recent block. This way, when requested later, React Query will use the cached value instead of performing the procedure call again.
  3.	Redirect to `/block/[neighbor-block-number]`, where `neighbor-block-number` is the block number returned by the previous getBlockNeighbor call. When the page is rendered, tRPC will use the data from the cache we populated previously.

See [server side helpers](https://trpc.io/docs/client/nextjs/server-side-helpers) for performing trpc calls on the server side

See [`getQueryKey`] (https://trpc.io/docs/client/react/getQueryKey) to check how to populate React Query's cache entries

apps/web/src/components/content.tsx
```diff
@@ -25,6 +25,21 @@ function resolveApiUrl(): string {
   return `https://api.${env.NEXT_PUBLIC_NETWORK_NAME}.blobscan.com`;
 }
 
+type Network = typeof env.NEXT_PUBLIC_NETWORK_NAME;
+
+const NETWORKS_FIRST_BLOB_NUMBER: Record<Network, number> = {
```
PabloCastellano, line 30: Maybe we can have something like **getNetworkDencunForkSlot** but for blocks?

https://github.com/Blobscan/blobscan/blob/03fb6b0d3291a85e80cbdab6cb497b782b17e7e8/apps/rest-api-server/src/utils.ts#L3-L20

PabloCastellano (commented):
I miss some kind of effect in the buttons so that they change either on hover or clicked. Overall the new implementation using cache looks good!

Reuse the component we have already.

PabloCastellano (approved):

PR #481, opened 2024-08-05
Make SWARM_DEFERRED_UPLOAD default to true

PabloCastellano:
### Checklist

### Description

Change default value for the SWARM_DEFERRED_UPLOAD variable.

We had some problems in the past with the swarm bee client and deferred uploads, so we set it to False. It looks like these issues are gone since the bee v2.0.0 release so it is safe to set it back to true by default. It is also required in order to keep the swarm storage in sync with the blobs that are processed.


PJColombo (approved):
LGTM!

PR #486, opened 2024-08-05
refactor: button component - improve props, variants, sizing, icons, …

luis-herasme:
### Description
1. The current situation with button props could be better. We manually add only a subset of the possible props to the button component. With this change, all the possible button attributes can be passed to the button Component.

2. Adding new variants used to be challenging. Previously, we had to define two types to add a new variant: the variant type key and the variant object. Then, we would add a new property to the button props type. Now, to add a new variant, we only need to add a new value to the CVA object.

3. Adding a tailwind class to the `className` already defined inside the component can cause uncertainty about which `className` will be applied. To solve this now, we are using tailwind-merge, which merges tailwind classes and keeps the last `className` added. With this, we can guarantee that the `className` defined by the user will be the `className` used.
	
4. We previously had a size enum defined externally, which led to several issues. The enum did not have a complete definition to cover all cases. For instance, `2xl` and `xs` could be passed to the size prop, but we did not account for them. Additionally, we only handled `xl` when it was an icon. Our sizing logic became very complex because we had to consider the icons, even though currently, all icons are the same size.

5. Enable tailwind intellisense inside variant definition using `.vscode/setting.json` (VSCode only config)


.changeset/mighty-balloons-camp.md
```diff
@@ -0,0 +1,5 @@
+---
+"@blobscan/web": minor
+---
+
+refactor: Improved button component - props, variants and sizing logic"
```
xFJA, line 5:
```suggestion
refactor: Improved button component - props, variants and sizing logic
```
typo

xFJA (approved):
Good job! :muscle:

PJColombo (changes requested):
Great job!.

I really like idea of  using of CVA for defining component variants and the integration of `tailwind-merge` to effectively combine Tailwind classes. 

However, I have some concerns about the introduction of `md-icon` as it seems to disrupt the overall sizing logic, making it feel more like an ad hoc solution. To address this, I suggest abstracting the icon variant logic into a separate `IconButton` component. This component would manage its own size definitions (e.g., sm =`h-3 w-3`, md = ` h-5 w-5,` etc.) which would then be passed down to the `children` icon prop. This way, we wouldn’t need to define the size explicitly in each component, as we have been doing so far. The provided icon children would be wrapped within a `div` using the following Tailwind classes: `flex items-center justify-center`.

PJColombo (approved):
Pushed a last refactoring change to the `IconButton` component

LGTM, feel free to merge

Issue #465, opened 2024-07-31
Create rollup filter


PJColombo:
It should be a dropdown component that shows a list of rollup elements comprised of:

1. The rollup's logo
2. The rollup's name

When no element is selected it should display "Rollup" inside the dropdown

### Notes

- We should do #490 changes in `Dropdown` component before we start his task

PR #491, opened 2024-08-06
refactor: Update `Dropdown` items by new Option type  

xFJA:
### Checklist

### Description

`Dropdown` component `items` have been renamed to `options` and the type has been changed to new `Option` to allow adding new elements to the option display further.

#### Motivation and Context
We need to add an icon as a prefix in #465 in each option of the Dropdown. With this new approach, we can easily add a new prop to `Option` type and use it in the `Option` subcomponent


apps/web/src/components/Dropdown/components/Option.tsx
```diff
@@ -0,0 +1,28 @@
+import { ListboxOption } from "@headlessui/react";
+
+import type { Option as OptionProps } from "../Dropdown";
+
+export const Option: React.FC<OptionProps> = function ({ label, value }) {
+  return (
+    <ListboxOption
+      className={({ active }) =>
```
luis-herasme, line 8: `active` is deprecated.ts(6385)
listbox.d.ts(62, 9): The declaration was marked as deprecated here

```typescript
type OptionRenderPropArg = {
    /** @deprecated use `focus` instead */
    active: boolean;
    focus: boolean;
    selected: boolean;
    disabled: boolean;
    selectedOption: boolean;
};
```

apps/web/src/components/Dropdown/Dropdown.tsx
```diff
@@ -34,40 +46,20 @@ export const Dropdown: React.FC<DropdownProps> = function ({
...
-                      }`}
-                    >
-                      {item}
-                    </span>
-                  </>
-                )}
-              </Listbox.Option>
+          <ListboxOptions className="absolute z-50 mt-1 max-h-60 w-full overflow-auto rounded-md bg-controlBackground-light py-1 text-base shadow-lg ring-1 ring-black ring-opacity-5 focus:outline-none dark:bg-controlBackground-dark sm:text-sm">
+            {options.map((option, id) => (
+              <Fragment key={id}>
```
luis-herasme, line 58: We can replace this:
```tsx
<Fragment key={id}>
  <Option {...option} />
</Fragment>
```
With this:
```tsx
  <Option {...option} key={id} />
```

apps/web/src/components/Dropdown/Dropdown.tsx
```diff
@@ -34,40 +46,20 @@ export const Dropdown: React.FC<DropdownProps> = function ({
               aria-hidden="true"
             />
           </span>
-        </Listbox.Button>
+        </ListboxButton>
         <Transition
           as={Fragment}
```
luis-herasme, line 51: This is the default value, so we do not have to set it:
image

https://headlessui.com/react/transition

apps/web/src/components/Dropdown/components/Option.tsx
```diff
@@ -0,0 +1,28 @@
...
+          <span
+            className={`block truncate text-sm ${selected ? "font-bold" : ""}`}
+          >
+            {label}
+          </span>
+        </div>
+      )}
+    </ListboxOption>
+  );
+};
```
luis-herasme, line 28: `active` is deprecated.ts(6385)
listbox.d.ts(62, 9): The declaration was marked as deprecated here

```typescript
type OptionRenderPropArg = {
    /** @deprecated use `focus` instead */
    active: boolean;
    focus: boolean;
    selected: boolean;
    disabled: boolean;
    selectedOption: boolean;
};
```
Here, we replace `active` with `focus` but also modify the content of the `ListboxOption` to accept `children` instead of a label. This still works if you pass a `string` but will work with custom components, too. For example, a component that has a label and an icon. This way, we do not have to modify the dropdown itself to add a dropdown item with an icon.
```suggestion
export const Option: React.FC<OptionProps> = function ({ children, value }) {
  return (
    <ListboxOption
      className={({ focus, selected }) =>
        `flex cursor-pointer select-none items-center px-4 py-2 text-sm
          ${selected ? "font-bold" : "font-normal"}
          ${
            focus
              ? "bg-controlActive-light dark:bg-controlActive-dark dark:text-content-dark"
              : "text-contentSecondary-light dark:text-contentSecondary-dark"
          }
        `
      }
      value={value}
    >
      {children}
    </ListboxOption>
  );
};
```

apps/web/src/components/Dropdown/Dropdown.tsx
xFJA: Done :+1:

apps/web/src/components/Dropdown/Dropdown.tsx
xFJA: Removed :+1:

apps/web/src/components/Dropdown/components/Option.tsx
xFJA: As we agreed, we are going to keep this approach to keep consistency between different variations of the `Option`: with an icon as a prefix, with another component as a suffix... following a slotted approach further if we want to add new elements and keep all `Dropdown` scenarios homogeneous

Regarding the `active` deprecated prop, it has been replaced by `focus` :+1:

apps/web/src/components/Dropdown/Dropdown.tsx
PJColombo, line 1: Let's rename this to `index.tsx`

apps/web/src/components/Dropdown/Dropdown.tsx
```diff
@@ -0,0 +1,63 @@
...
+  ListboxOptions,
+  Transition,
+} from "@headlessui/react";
+import { ChevronUpDownIcon } from "@heroicons/react/24/outline";
+
+import { Option } from "./components/Option";
+
+export interface Option {
+  value: string | number;
+  label: string;
```
PJColombo, line 13: I’d prefer to generalize the `label` prop to accept a `ReactNode` instead

PJColombo, line 13: Also, let's make it optional and render the `value` when it's not set

apps/web/src/components/Dropdown/Dropdown.tsx
```diff
@@ -0,0 +1,63 @@
...
+}
+
+export interface DropdownProps {
+  options: Option[];
+  selected: Option;
```
PJColombo, line 18: I'd revert this to the previous prop and accept the index of one of the elements pass in the `options` props.

I think it was simpler to set

apps/web/src/components/Dropdown/Dropdown.tsx
```diff
@@ -0,0 +1,63 @@
...
+  width?: string;
+  onChange(newOption: Option): void;
```
PJColombo, line 20: `onChange` can be as follows:

`onChange(index, options)` where `index` is the position in the array of the selected element and `options` is props array

apps/web/src/components/Dropdown/components/Option.tsx
PJColombo, line 1: Can we add related components directly to the main component’s folder, rather than creating a nested `components` folder?.

This approach would maintain consistency with the rest of the app's components

PJColombo (changes requested):
Nice work. I like the refactoring.

I left some request changes

apps/web/src/components/Dropdown/components/Option.tsx
xFJA: I created this subcomponents directory because:
- You will have the main `/components` directory more clear and easy to navigate
- You are only going to use this component in the `Dropdown` scope and not in another component. I would only move this to the main component's root folder if it's going to be used by another component

apps/web/src/components/Dropdown/Dropdown.tsx
xFJA: I think the current solution is simpler and matches with most `selects/dropdowns` implementations along well-known libraries: when this event has been triggered, you should pass the new value of the HTML element. If you pass the `index` and the whole `options` array, you have two drawbacks:
- You will have to look for the value always when you change the value 
- You already have the `options` array when you pass it as a prop to the component so you would be sending redundant information

apps/web/src/components/Dropdown/Dropdown.tsx
xFJA: Done :+1:

apps/web/src/components/Dropdown/components/Option.tsx
```diff
@@ -0,0 +1,28 @@
+import { ListboxOption } from "@headlessui/react";
+
+import type { Option as OptionProps } from "..";
+
+export const Option: React.FC<OptionProps> = function (props) {
+  return (
+    <ListboxOption
+      className={({ focus }) =>
+        `relative cursor-pointer select-none py-2 pl-4 pr-4 font-normal ${
```
luis-herasme, line 9:
```suggestion
        `relative cursor-pointer select-none py-2 px-4 ${
```

apps/web/src/components/Dropdown/components/Option.tsx
xFJA: Done :+1:

apps/web/src/components/Dropdown/Dropdown.tsx
xFJA: Done :+1:

apps/web/src/components/Dropdown/components/Option.tsx
xFJA: Done :+1:

luis-herasme (approved):

apps/web/src/pages/blob/[hash].tsx
```diff
@@ -49,6 +50,9 @@ const Blob: NextPage = function () {
     ...(blob && decoder ? ["Decoded" as BlobViewMode] : []),
     ...DEFAULT_BLOB_VIEW_MODES,
   ];
+  const blobViewModesOptions: DropdownProps["options"] = blobViewModes.map(
+    (blobViewMode) => ({ label: blobViewMode, value: blobViewMode })
```
Comment (PJColombo) [inline apps/web/src/pages/blob/[hash].tsx:54]:
```suggestion
    (blobViewMode) => ({ value: blobViewMode })
```

apps/web/src/components/PaginatedTable/PaginatedTable.tsx
```diff
@@ -130,8 +138,8 @@ export const PaginatedTable: FC<PaginatedTableProps> = function ({
             <div className="flex items-center justify-start gap-2">
               Displayed items:
               <Dropdown
-                items={PAGE_SIZES}
-                selected={pageSize}
+                options={PAGE_SIZES_OPTIONS}
+                selected={{ value: pageSize, label: pageSize.toString() }}
```
PJColombo, line 142:
```suggestion
                selected={{ value: pageSize }}
```

apps/web/src/components/Layouts/PaginatedListLayout.tsx
```diff
@@ -110,8 +118,8 @@ export const PaginatedListLayout: FC<PaginatedListLayoutProps> = function ({
               <div className="flex items-center justify-start gap-2">
                 Displayed items:
                 <Dropdown
-                  items={PAGE_SIZES}
-                  selected={pageSize}
+                  options={PAGE_SIZES_OPTIONS}
+                  selected={{ value: pageSize, label: pageSize.toString() }}
```
PJColombo, line 122:
```suggestion
                  selected={{ value: pageSize }}
```

apps/web/src/components/Layouts/PaginatedListLayout.tsx
```diff
@@ -20,7 +20,12 @@ export type PaginatedListLayoutProps = {
   emptyState?: ReactNode;
 };
 
-const PAGE_SIZES = [10, 25, 50, 100];
+const PAGE_SIZES_OPTIONS: DropdownProps["options"] = [
+  { value: 10, label: "10" },
+  { value: 25, label: "25" },
+  { value: 50, label: "50" },
+  { value: 100, label: "100" },
+];
```
PJColombo, line 28:
```suggestion
const PAGE_SIZES_OPTIONS: DropdownProps["options"] = [
  { value: 10 },
  { value: 25 },
  { value: 50 },
  { value: 100 },
];
```

PJColombo (changes requested):
Now that `label` is optional we can make the following changes

apps/web/src/components/Dropdown/Dropdown.tsx
PJColombo: Ok, let's continue with this approach

PJColombo:
LGTM.

Feel free to merge once we resolve previous comments

apps/web/src/pages/blob/[hash].tsx
Comment (xFJA) [reply apps/web/src/pages/blob/[hash].tsx]: Done :+1:

apps/web/src/components/PaginatedTable/PaginatedTable.tsx
xFJA: Done :+1:

apps/web/src/components/Layouts/PaginatedListLayout.tsx
xFJA: Done :+1:

apps/web/src/components/Layouts/PaginatedListLayout.tsx
xFJA: Done :+1:

PR #493, opened 2024-08-07
fix(web): Correct formatWei precision for large decimal values

### LINKED ISSUE #485: Fix precision error
We are using 17 decimals instead of 18.

image
https://blobscan.com/tx/0x05f4bfb01bbb4263d8b015f9a11ab9de5577c36c5d68f4ffb48ec7dfafe2cb54

luis-herasme:
### Description
1. Refactored the `eth-format` utils into a new package.
2. Corrected `formatWei` precision for large decimal values.
3. Added tests for `formatWei`.
4. Corrected `formatWei` error with negative numbers. 

### Context
The original problem happened because, inside the `formatWei` function, we were converting back and forth between `Number` and `String` and handling numbers higher than the [MAX_SAFE_INTEGER](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/MAX_SAFE_INTEGER).

For example, inside `formatWei`, we used the `formatNumber` function, which converts the input to the number before calling the formatter.

The function `formatWei` had too many parameters and performed too many tasks, which made it difficult to test. The function used to take an options object with the following values: `toUnit`, `compact`, and `displayUnit`. While refactoring the code, I noticed that the `displayUnit` was always set to the default value, so we could remove this parameter. Also, since the `compact` parameter was only used in one place, we can remove it as well.

Another observation was that `formatWei` takes `number` | `bigint`. This complicated the conversion between units because we had to handle the decimal part of the number. However, since we won't have a case with fractional `wei`, we didn't need to add that extra complexity. Now, `formatWei` only takes `bigint` as input.			

After that, I was able to add some tests and realized that the previous implementation did not support negative numbers, so I also fixed that.

#### Update
Added `prettyFormatWei`. The difference between this function and `formatWei` is that this function does not ensure that the full precision of the input is preserved. Instead, this function provides a more human-readable representation of the value.

### Related Issue
Closes #485

### DIFF
--- a/.changeset/stupid-plants-beam.md
+++ b/.changeset/stupid-plants-beam.md
@@ -1,5 +1,5 @@
 ---
-"@blobscan/web": patch
+"@blobscan/web": minor
 ---
 
 Corrected formatWei precision for large decimal values
--- a/.changeset/stupid-plants-beam.md
+++ b/.changeset/stupid-plants-beam.md
@@ -2,4 +2,4 @@
 "@blobscan/web": patch
 ---
 
-fix(web): Correct formatWei precision for large decimal values
+Corrected formatWei precision for large decimal values
--- a/.changeset/stupid-plants-beam.md
+++ b/.changeset/stupid-plants-beam.md
@@ -1,0 +1,5 @@
+---
+"@blobscan/web": patch
+---
+
+fix(web): Correct formatWei precision for large decimal values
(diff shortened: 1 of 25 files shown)


PabloCastellano (commented):
Can you add a test case to make sure it won't happen again?

.changeset/stupid-plants-beam.md
```diff
@@ -0,0 +1,5 @@
+---
+"@blobscan/web": patch
+---
+
+fix(web): Correct formatWei precision for large decimal values
```
xFJA, line 5:
```suggestion
Corrected formatWei precision for large decimal values
```
_[picky]_

I've checked other changeset entries:
- the prefix is not added
- we use the past verb

xFJA (approved):

PabloCastellano (approved):
Since you are removing existing functions in the package, this can't be a patch release. Please change it to minor. Other than this, lgtm!

.changeset/stupid-plants-beam.md
```diff
@@ -0,0 +1,5 @@
+---
+"@blobscan/web": patch
```
PabloCastellano, line 2: should be minor at least

PabloCastellano:
Also remember to include this package in the docs at https://docs.blobscan.com/docs/codebase-overview

.changeset/stupid-plants-beam.md
luis-herasme: Done [a25a27d](https://github.com/Blobscan/blobscan/pull/493/commits/a25a27daa36b17b956d79be9fef9a051c551ec07)

PR #496, opened 2024-08-08
feat: create `Rollup` filter & `FilterPanel`

### LINKED ISSUE #434: Add filters to blobs, blocks and transactions pages

- [x] #465
- [ ] #466
- [ ] #467

Mockup:

image

### LINKED ISSUE #465: Create rollup filter

xFJA:
### Checklist

### Description

- Add to `Dropdown` option a `prefix ` prop to be used as `slot` -> This scenario passes a `RollupIcon`
- Create `FilterPanel` to be used as a base for more filters
- Add a `rollup` dropdown to be able to filter in the `blobs`, `blocks`, and `transactions` view tables by `rollup`
- Fix pages number in `<PaginatedTable />`

### Related Issue 

- Closes #465 
- Closes #434 

### Screenshots

image


apps/web/src/components/PaginatedTable/components/RollupFilter.tsx
```diff
@@ -0,0 +1,53 @@
+import type { FC } from "react";
+
+import { Dropdown } from "~/components/Dropdown/Dropdown";
+import type { DropdownProps } from "~/components/Dropdown/Dropdown";
+import { RollupIcon } from "~/components/RollupIcon";
+import type { Rollup } from "~/types";
+
+const rollups: Rollup[] = [
```
PabloCastellano, line 8: I'd rather use the existing Rollups enum instead of redefining the list of rollups here.

PJColombo: +1

apps/web/src/components/PaginatedTable/components/FilterPanel.tsx
PJColombo, line 1: Let's move this up to the `/components` folder

PJColombo, line 1: I’d prefer to name this component to `Filters`.`FiltersPanel` might suggest it’s just the panel, whereas it actually represents the entire card containing all the filters.

PJColombo, line 1: Also, Can we add related components directly to the main component’s folder, rather than creating a nested `components` folder?. This approach would maintain consistency with the rest of the app's components

apps/web/src/components/PaginatedTable/PaginatedTable.tsx
PJColombo, line 1: Let’s rename this to index.tsx to avoid redundancy when importing, so we don’t end up with `PaginatedTable/PaginatedTable.tsx`.

apps/web/src/components/RollupIcon.tsx
PJColombo, line 1: No need to have this if we use `RollupBadge`

.changeset/smooth-trains-fly.md
```diff
@@ -0,0 +1,5 @@
+---
+"@blobscan/web": minor
+---
+
+Added Rollup filter to PaginatedTable filter panel
```
PJColombo, line 5:
```suggestion
Added rollup filter to blobs, blocks and transactions views
```

PJColombo (changes requested):
Nice work!

I left a few change requests.

I also think it would be better to trigger a redirection to a route with filter query parameters based on the configured filters, rather than re-executing the TRPC fetch call.

The "Filter" button would behave as a link with all the filter query params set

This approach allows the user to share the url with specific filters and maintains consistency with the rest of the routing in the app.

PabloCastellano:
> I also think it would be better to trigger a redirection to a route with filter query parameters based on the configured filters, rather than re-executing the TRPC fetch call.

Suggestion: Instead of using the rollup name as a query parameter, let's use the address that sends the blobs. This will serve as an escape hatch to filter out rollups we haven't identified yet.

apps/web/src/components/PaginatedTable/components/FilterPanel.tsx
xFJA: Let's keep this thread opened til we agreed on https://github.com/Blobscan/blobscan/pull/491#discussion_r1721688967 that is the same issue!

apps/web/src/components/PaginatedTable/components/RollupFilter.tsx
```diff
@@ -0,0 +1,53 @@
...
+];
+
+const ROLLUP_OPTIONS: DropdownProps["options"] = rollups.map((rollup) => {
+  const icon = <RollupIcon rollup={rollup} />;
+  return {
+    value: rollup,
+    label: rollup.charAt(0).toUpperCase() + rollup.slice(1),
+    prefix: icon === null ? <div className="h-4 w-4"></div> : icon,
+  };
+});
```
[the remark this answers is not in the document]
xFJA: I think we should keep the `Option` implementation for `RollupFilter` decoupled from the `RollupBadge` because they are different components:

The badge is used here 
image
where a different background/text color is applied for all the rollups. This will cause a dropdown component full of different options styles.

Moreover, this implementation follows the [ new refactor](https://github.com/Blobscan/blobscan/pull/491) for the `Dropdown` component where we want to maintain consistency between all the dropdown currencies in the application. If we pass a whole component like `RollupBadge`, we'll have non-homogeneous implementations

apps/web/src/components/PaginatedTable/components/FilterPanel.tsx
xFJA: Let's keep this thread opened til we agreed on https://github.com/Blobscan/blobscan/pull/491#discussion_r1721688967 that is the same issue!

apps/web/src/components/RollupIcon.tsx
xFJA: I will keep this thread opened til we solve https://github.com/Blobscan/blobscan/pull/496#discussion_r1721772567 as they are related

apps/web/src/components/PaginatedTable/components/FilterPanel.tsx
xFJA: Done

apps/web/src/components/PaginatedTable/PaginatedTable.tsx
xFJA: Done

apps/web/src/components/PaginatedTable/components/RollupFilter.tsx
xFJA: [Done](https://github.com/Blobscan/blobscan/pull/496/commits/3444ceb62a8939b49946019a41f8b8d761168fb0) by @luis-herasme in

apps/web/src/components/PaginatedTable/components/FilterPanel.tsx
xFJA: Done :+1:

.changeset/smooth-trains-fly.md
xFJA: Done

PabloCastellano (commented):
Some more feedback:

- Right now if I select Arbitrum, I can't go back to 'all rollups' (I have to refresh the page so that the dropdown is reset)
- The trpc call times out in the vercel preview. Is there a performance issue here? :thinking: 

image

xFJA:
> Some more feedback:
> 
> * Right now if I select Arbitrum, I can't go back to 'all rollups' (I have to refresh the page so that the dropdown is reset)
> * The trpc call to filter by rollup times out in the vercel preview. Is there a performance issue here? 🤔
> 
> image

As we have agreed, a new `Clear` button have been added that will reset all filters :+1:

apps/web/src/components/Filters/RollupFilter.tsx
```diff
@@ -0,0 +1,32 @@
...
+}) {
+  return (
+    <>
+      <Dropdown
+        selected={selected}
+        options={Object.values(Rollups).map((rollup) => ({
+          value: rollup.toLowerCase(),
+          label: capitalize(rollup),
+          prefix: <RollupIcon rollup={rollup.toLowerCase() as Rollup} />,
+        }))}
```
PJColombo, line 25: I forgot to mention this in the Dropdown refactor PR: I believe the dropdown’s `prefix` prop is unnecessary now that we can pass a React node as the label. We can simply include the prefix directly in the label component.

Could we instead have a label that includes both the `RollupIcon` component and the capitalized rollup name? This way, the selected rollup would be rendered with the icon, as opposed to the current implementation where only the text appears.

image

apps/web/src/components/PaginatedTable/components/RollupFilter.tsx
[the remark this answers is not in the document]
PJColombo: Let's keep it like it is now

apps/web/src/components/PaginatedTable/index.tsx
```diff
@@ -120,6 +126,9 @@ export const PaginatedTable: FC<PaginatedTableProps> = function ({
         emptyState={DEFAULT_TABLE_EMPTY_STATE}
       >
         <div className="flex flex-col gap-6">
+          <div className="w-1/2">
+            <Filters />
+          </div>
           <Table
```
PJColombo, line 132: Can we move the `Filters` out of  the `PaginatedTable` component?

Let's have them wrapped within their own `Card` component on top of the table card

PJColombo (changes requested):
Left some comments.

Also, I noticed that when the user navigates to a route with filter query params set, the rollup filter component doesn't have it set as well 

For example, If I go [here](https://blobscan-staging-ptcdiirs5-blossom-labs.vercel.app/txs?rollup=camp) you can see that the rollup dropdown doesn't have the camp rollup selected

apps/web/src/components/Filters/RollupFilter.tsx
PJColombo: I'd use drop the `prefix` prop as well from the dropdown component

apps/web/src/components/PaginatedTable/index.tsx
PJColombo: It should also take the whole width

apps/web/src/components/Filters/RollupFilter.tsx
```diff
@@ -0,0 +1,49 @@
...
+export const RollupFilter: FC<RollupFilterProps> = function ({
+  onChange,
+  selected,
+}) {
+  return (
+    <Dropdown
+      selected={selected}
+      options={ROLLUP_OPTIONS}
+      onChange={onChange}
+      placeholder="Rollup"
```
PabloCastellano, line 44:
```suggestion
      placeholder="All rollups"
```

luis-herasme (approved):

PR #497, opened 2024-08-10
feat: create `Timestamp` filter

### LINKED ISSUE #434: Add filters to blobs, blocks and transactions pages

### LINKED ISSUE #466: Create a timestamp filter
Use the following [datepicker](https://react-tailwindcss-datepicker.vercel.app)

The filter should be comprised of two joined text fields separated by a vertical bar. When click either of them the date picker should be displayed. 

They should display the following hints: "Start date" and "End date"

xFJA:
### Checklist

### Description

- Add to `Dropdown` a custom `height` prop to keep all inputs inside `FilterPanel` with same height
- Add `fullWidth` prop to `Button` to be used in small screens layouts
- Install library to have a Tailwind React `DatePicker` component
- Add a `timestamp` dropdown to be able to filter in the `blobs`, `blocks`, and `transactions` view tables by start and end date

### Related Issue 

- Closes #466  
- Closes #434 

### Screenshots 

image


PabloCastellano (commented):
I tried this PR in the Vercel preview, and here are some notes:

- The filter doesn't actually seem to work (and the Filter button is always gray (is it disabled?)).
- Is the number of blob transactions updated after the filters are applied? From your screenshot, it looks like it didn't get updated accordingly.

xFJA:
> I tried this PR in the Vercel preview, and here are some notes:
> 
> * The filter doesn't actually seem to work (I select dates and nothing happens, also the Filter button is always gray (is it disabled?)).
> * Is the number of blob transactions updated after the filters are applied? From your screenshot, it looks like it didn't get updated accordingly.

The branch was missing a `git rebase` because many files in the target branch have been refactored! You can check now :+1:

PabloCastellano (commented):
I've tried again and picking the dates won't let me click the Filter button.

We should also allow users to filter by date range, without selecting any particular rollup.

xFJA:
> I've tried again and picking the dates won't let me click the Filter button.
> 
> We should also allow users to filter by date range, without selecting any particular rollup.

The conditions to enable the filter button has been reduced to at least one filled. Now you can filter using the timestamp without selecting the rollup

PabloCastellano (commented):
Some bugs still present:

- The number of blob transactions is not updated after the filters are applied
- It is impossible to remove the date range once selected due to the button being disabled when I remove the dates.

xFJA:
> Some bugs still present:
> 
> * The number of blob transactions is not updated after the filters are applied
> * It is impossible to remove the date range once selected due to the button being disabled when I remove the dates.

Added a new statement to check whether all the filters have been reset individually. If `true`, the table is reset to all items automatically :+1:

PJColombo (approved):
Merged rollup filter latest changes and aligned the styles of the dropdown and date picker

luis-herasme (approved):

apps/web/src/hooks/useQueryParams.ts
```diff
@@ -6,10 +6,12 @@ import { Rollup } from "@blobscan/api/enums";
 import type { Rollup as LowercaseRollup } from "~/types";
 
 type QueryParams = {
+  from?: string;
   p: number;
   ps: number;
   rollup?: LowercaseRollup | "null";
```
xFJA, line 12: I think the `rollup` shouldn´t be a query param as we use the `from` mapped address associated instead

apps/web/src/utils/date.ts
```diff
@@ -6,6 +6,14 @@ export function normalizeTimestamp(timestamp: number | Date | string) {
     : dayjs(timestamp);
 }
 
+export function getISODate(date: dayjs.Dayjs | Date | string) {
```
xFJA, line 9: should we add UTs here?

packages/api/src/app-router.ts
```diff
@@ -32,6 +33,7 @@ export const appRouter = t.router({
     .input(z.void())
     .output(z.string())
     .query(() => "yay!"),
+  getRollups,
```
xFJA, line 36: These changes were removed in the rollup filter refactor. This should be an error in managing conflicts after many merges

packages/api/src/getRollups.ts
```diff
@@ -0,0 +1,9 @@
+import { Rollup } from "@blobscan/db";
```
xFJA, line 1: These changes were removed in the rollup filter refactor. This should be an error in managing conflicts after many merges

xFJA (commented):
It´s seems the background color overriding for the date range picker component has dissappeared in the refactor because the colors now mismatch: 

image

apps/web/src/hooks/useQueryParams.ts
PJColombo: Actually, we still need it in order to filter non-rollup elements (`rollup=null`)

PJColombo:
> It´s seems the background color overriding for the date range picker component has dissappeared in the refactor because the colors now mismatch:
> 
> image

The colours look fine to me

image

apps/web/src/utils/date.ts
PJColombo: Let's do component and e2e testing for the web app instead

PR #502, opened 2024-08-13
ci: Fix prisma version mismatch between @prisma and @prisma-client

PabloCastellano:
### Checklist

### Description

Fix prisma version mismatch between @prisma and @prisma-client

The warning was being shown in the build logs:

```
@blobscan/db:db:generate: warn Versions of prisma@5.5.2 and @prisma-client@5.17.0 don't match.
@blobscan/db:db:generate: This might lead to unexpected behavior.
@blobscan/db:db:generate: Please make sure they have the same version.
```

### DIFF
--- a/packages/db/package.json
+++ b/packages/db/package.json
@@ -29,7 +29,7 @@
     "@blobscan/logger": "workspace:^0.1.1",
     "@blobscan/open-telemetry": "workspace:^0.0.8",
     "@prisma/client": "^5.5.2",
-    "prisma": "5.5.2"
+    "prisma": "5.17.0"
   },
   "devDependencies": {
     "@faker-js/faker": "^8.0.2",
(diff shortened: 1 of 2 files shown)


packages/db/package.json
```diff
@@ -29,7 +29,7 @@
     "@blobscan/logger": "workspace:^0.1.1",
     "@blobscan/open-telemetry": "workspace:^0.0.8",
     "@prisma/client": "^5.5.2",
-    "prisma": "5.5.2"
+    "prisma": "5.17.0"
```
PJColombo, line 32: Why not use `^` operator so it installs latest minor update for both `prisma` and `@prisma/client` packages?

```suggestion
    "prisma": "^5.5.2"
```

PJColombo (commented):
Why not use `^` operator so it installs latest minor update for both `prisma`  and `@prisma/client` packages?

packages/db/package.json
PabloCastellano: Yes that looks better. I took the opportunity to bump them to the latest version available

luis-herasme (approved):

PR #504, opened 2024-08-14
Best gas unit

### LINKED ISSUE #500: Find best unit for Blob gas price
Once https://github.com/Blobscan/blobscan/pull/493 gets merged, use `findBestUnit  from "@blobscan/eth-units";` for the new Blob gas price in the header. It should display `1 wei` instead of `0.000000001 Gwei`.

* [ ] Apply to Header

image

* [ ] Apply to Block / Transaction details views. Leave in both Gwei and Wei?

image

luis-herasme:
### Description
Adjust blob gas price components for a better unit representation

### Related Issue
Closes #500

### DIFF
--- a/.changeset/hip-maps-agree.md
+++ b/.changeset/hip-maps-agree.md
@@ -1,0 +1,5 @@
+---
+"@blobscan/web": minor
+---
+
+Optimized blob gas price components to use most appropriate unit
--- a/.changeset/hip-maps-agree.md
+++ b/.changeset/hip-maps-agree.md
@@ -1,5 +1,5 @@
 ---
-"@blobscan/web": minor
+"@blobscan/web": patch
 ---
 
-Updated latest blob gas price metric to display the most suitable unit, replacing the previously  fixed `Gwei` format
+Updated latest blob gas price metric to display the most suitable unit, replacing the previously fixed `Gwei` format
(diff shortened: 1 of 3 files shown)


.changeset/hip-maps-agree.md
```diff
@@ -0,0 +1,5 @@
+---
+"@blobscan/web": minor
```
PJColombo, line 2: Rather than a `minor` change I would say its a patch one as we're changing an existing component

.changeset/hip-maps-agree.md
```diff
@@ -0,0 +1,5 @@
...
+---
+
+Optimized blob gas price components to use most appropriate unit
```
PJColombo, line 5:
```suggestion
Updated latest blob gas price metric to display the most suitable unit, replacing the previously  fixed `Gwei` format
```

PJColombo (changes requested):
LGTM!

Left a few comments regarding the changeset description

.changeset/hip-maps-agree.md
luis-herasme: Done [1c108c6](https://github.com/Blobscan/blobscan/pull/504/commits/1c108c6ea563872fee4b5f0b98792e3de307d3d6)

PJColombo (approved):
LGTM!

PabloCastellano (approved):
lgtm

image

PR #507, opened 2024-08-19
Prevent tooltip overflow

luis-herasme:
### Description
The Tooltip component now includes extra logic to prevent it from overflowing.

### Screenshots:
Before:
image

After:
image


PJColombo (changes requested):
Could we maybe use the [Popover](https://headlessui.com/v1/react/popover) element from headlessui?

The tooltip logic is becoming increasingly complex, and rather than reinventing the wheel, it would be more efficient to leverage a battle-tested library that we’re already using in other parts of the codebase.

PJColombo:
Additionally, after doing some research, it appears that Headless UI doesn’t natively support opening the Popover on hover. However, I found a [workaround](https://github.com/tailwindlabs/headlessui/discussions/425#discussioncomment-5117609)

apps/web/src/components/CopyToClipboard.tsx
```diff
@@ -15,29 +14,31 @@ export function CopyToClipboard({
   value,
 }: CopyToClipboardProps) {
   const [isCopied, setIsCopied] = useState(false);
-  const buttonRef = useRef<HTMLButtonElement>(null);
-  const isHovered = useHover(buttonRef);
-  useEffect(() => setIsCopied(false), [isHovered]);
 
+  // TODO: Use Button component
```
PabloCastellano, line 18: does this comment still make sense?

apps/web/src/components/Tooltip.tsx
PJColombo, line 1: Can we add some extra lateral padding to the tooltip?

image

apps/web/src/components/Tooltip.tsx
```diff
@@ -1,21 +1,173 @@
...
+  useHover,
+  useFocus,
+  useDismiss,
+  useRole,
+  useInteractions,
+  FloatingPortal,
+  FloatingArrow,
+  arrow,
+  useMergeRefs,
+} from "@floating-ui/react";
```
PJColombo, line 28: Where are we importing this dep from? It's not in the `package.json`

PJColombo (changes requested):
Nice work

I left some comments

apps/web/src/components/Tooltip.tsx
luis-herasme: Done https://github.com/Blobscan/blobscan/pull/507/commits/3561fd20ab6a73bde9d9ed820eded48a93b128d7

apps/web/src/components/Tooltip.tsx
luis-herasme: Done https://github.com/Blobscan/blobscan/pull/507/commits/04331adea2d153ab27fca1f57ff6d90bf60220fb

packages/api/package.json
```diff
@@ -27,7 +27,8 @@
     "jsonwebtoken": "^9.0.0",
     "superjson": "1.9.1",
     "trpc-openapi": "^1.2.0",
-    "viem": "^2.17.4"
+    "viem": "^2.17.4",
+    "@floating-ui/react": "*"
```
PJColombo, line 31: Let’s use `pnpm add` to install the package so that it adds a locked version, allowing only non-breaking updates.

Using `*` is not recommended, as it installs any new version, including those with potential breaking changes.

.changeset/tricky-actors-worry.md
```diff
@@ -0,0 +1,5 @@
+---
+"@blobscan/web": patch
+---
+
+The Tooltip component now includes extra logic to prevent it from overflowing.
```
PJColombo, line 5: Let's follow the standard and only use past tense verbs please

```suggestion
Added additional logic to the `Tooltip` component to prevent overflow.
```

apps/web/src/components/CopyToClipboard.tsx
```diff
@@ -15,29 +14,30 @@ export function CopyToClipboard({
...
-        <div className="whitespace-nowrap">{isCopied ? "Copied!" : label}</div>
-      </Tooltip>
-    </button>
+      <TooltipContent>{isCopied ? "Copied!" : label}</TooltipContent>
+      <TooltipTrigger
+        className="text-contentTertiary-light hover:text-link-light dark:text-contentTertiary-dark dark:hover:text-link-dark"
+        onClick={() => {
+          navigator.clipboard.writeText(value);
+          setIsCopied(true);
+        }}
```
PJColombo, line 32:
```suggestion
       onClick={() =>
          void navigator.clipboard
            .writeText(value)
            .then(() => {
              setIsCopied(true);
            })
            .catch((err) => {
              console.error("Failed to copy to clipboard: ", err);
            })
        }
```

PJColombo (changes requested):
Nice work

Left a couple of comments

packages/api/package.json
luis-herasme: Done [2ed8ac7](https://github.com/Blobscan/blobscan/pull/507/commits/2ed8ac787db049b63c5a841a25987f868df0975d)

PJColombo (approved):
LGTM

PR #515, opened 2024-08-21
Fix block stats

luis-herasme:


PabloCastellano (changes requested):
Still failing. Check https://blobscan-staging-git-fix-block-stats-blossom-labs.vercel.app/stats/block

image

PJColombo (changes requested):
We’re currently rounding the metrics instead of displaying the actual values.

To address this, I suggest updating the `convertWei` function to handle decimal values. As shown by the errors, there are cases where we encounter wei values with decimals, particularly when dealing with averages.

Additionally, I noticed that we’re always converting values to big integers before passing them to `convertWei`. Instead of doing this, we could let the function itself handle the normalization of input parameters, as it was done prior to the changes in PR #493.

PJColombo (approved):
LGTM

PabloCastellano (approved):
LGTM!

.changeset/light-wombats-try.md
```diff
@@ -0,0 +1,5 @@
+---
+"@blobscan/web": patch
+---
+
+Update DailyAvgBlobGasPriceChart to round average gas prices
```
PabloCastellano, line 5: Only missing thing is updating this description

PR #522, opened 2024-08-27
feat: create `BlockNumber` filter

### LINKED ISSUE #434: Add filters to blobs, blocks and transactions pages

### LINKED ISSUE #467: Create a block number filter
Create a block number filter that allows you to filter by a range of blocks.

The component should comprised of two numeric text inputs joined by a vertical bar.

Each text field should display the following hints: "Start block", "End block"

xFJA:
### Checklist

### Description

- Create a new `RangeInput` component to be able to select a range of numbers
- Add a `block number` range input to be able to filter in the `blobs`, `blocks`, and `transactions` view tables by start and end block number

### Related Issue 
- Closes #467 
- Closes #434 

### Screenshots

image


apps/web/src/components/Filters/index.tsx
```diff
@@ -3,43 +3,69 @@ import type { FC } from "react";
...
+  const [errors, setErrors] = useState<FiltersErrorsState>({
+    blockNumberRange: undefined,
+  });
+
   const router = useRouter();
 
   const breakpoint = useBreakpoint();
-  const fullWidth = breakpoint === "sm" || breakpoint === "default";
+  const fullWidth =
+    breakpoint === "md" || breakpoint === "sm" || breakpoint === "default";
```
PabloCastellano, line 47: Suggestion to improve readability: `const fullWidth = ["md", "sm", "default"].includes(breakpoint);`

apps/web/src/utils/validators.ts
```diff
@@ -0,0 +1,31 @@
...
+    return `Right range must be less than or equal to ${maxValue}.`;
+  }
+
+  if (leftRange && leftRange > maxValue) {
+    return `Left range must be less than ${maxValue}.`;
+  }
+
+  if (leftRange && rightRange && leftRange >= rightRange) {
+    return "The right-hand range cannot be less or equal to the left-hand range";
+  }
```
PabloCastellano, line 28: In general, I'd use "Start block" and "end block" instead of "left/right range" or "left/right-hand range"

apps/web/src/components/PaginatedTable/index.tsx
```diff
@@ -128,7 +129,7 @@ export const PaginatedTable: FC<PaginatedTableProps> = function ({
         emptyState={DEFAULT_TABLE_EMPTY_STATE}
       >
         <div className="flex flex-col gap-6">
-          <Filters />
+          {tableTopSlot && tableTopSlot}
```
PabloCastellano, line 132: Why did you do this change?

PabloCastellano (commented):
Left some questions about the code.

Also, 3 questions about the functionality:

* When I remove the start block number, the list is automatically filtered (without clicking the Filter button). Is this expected?
* Why are date range placeholders color red now?

image

* If I wanted to filter blobs from one block, I cannot do it right now, right?

image

apps/web/src/components/PaginatedTable/index.tsx
xFJA: The idea is to decouple the `Filter` component from the `PaginatedTable` component. It's a common strategy when building components to use the `slot` concept: you can pass by a prop whatever element (`ReactNode`) and will be rendered in this section: `tableTop` indicates the place where it will be rendered

apps/web/src/components/Filters/index.tsx
```diff
@@ -3,43 +3,69 @@ import type { FC } from "react";
...
+const DEFAULT_MIN_VALUE = BigInt(0);
+const DEFAULT_MAX_VALUE = BigInt(99999999);
+
+const INIT_STATE: FiltersState = {
   rollup: null,
   timestampRange: {
     startDate: null,
     endDate: null,
   },
+  blockNumberRange: [undefined, undefined],
```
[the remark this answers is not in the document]
xFJA: `timestampRange` uses a library for the `DatePicker` that forces to use the `null`, that the reason

apps/web/src/components/Inputs/NumberInput.tsx
```diff
@@ -0,0 +1,79 @@
...
+            return;
+          }
+
+          if (type === "uint" && newValueNumber < 0) {
+            return;
+          }
+        }
+
+        setInnerValue(newValue);
+        onChange(newValueNumber);
```
xFJA, line 73: Let´s extract all this logic into a func for clean coding

apps/web/src/components/Inputs/NumberRangeInput.tsx
```diff
@@ -0,0 +1,61 @@
...
+  end?: number;
+};
+
+export type NumberRangeInputProps = {
+  className?: InputHTMLAttributes<HTMLInputElement>["className"];
+  range: NumberRange;
+  type: NumberInputType;
+  onChange: (range: NumberRange) => void;
+  inputStartProps?: Omit<InputProps, "value" | "onChange" | "type">;
+  inputEndProps?: Omit<InputProps, "value" | "onChange" | "type">;
```
xFJA, line 18: We could create a type to avoid repeat it as both will have always the same one

apps/web/src/components/PaginatedTable/index.tsx
PJColombo: I’m reverting this because having both in the same card makes it look a bit too crowded. Let’s separate them into two different cards, with the filters on top of the paginated table.

apps/web/src/utils/validators.ts
PJColombo: I’m removing this validation logic since validation is now handled directly in the onChange event handler, avoiding the need to display an error message to the user.

apps/web/src/components/Filters/index.tsx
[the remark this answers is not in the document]
PJColombo: I’ve applied changes to normalize the initial value of the Filter component’s state. Using `null` for some inputs and `undefined` for others seemed too confusing.

PR #523, opened 2024-08-28
Separate Prisma enums to their own file

PJColombo:
### Checklist

- [ ] My change requires a documentation update, and I have done it.
- [ ] I have added tests to cover my changes.
- [x] I have filled out the description and linked the related issues.

### Description
It generates a Prisma enums file separate from the generated Prisma client and updates all enum imports in the codebase to reference this new file.

#### Motivation and Context (Optional)
Currently, it's not possible to import Prisma enums directly from the Prisma generated client on the client side due to this [issue](https://github.com/prisma/prisma/issues/13567#issue-1252911478)

This PR creates a separated enums file via a Prisma generator that allows us to circumvent the Prisma client and import the enums directly following this [gist](https://gist.github.com/juliusmarminge/b06a3e421117a56ba1fea54e2a4c0fcb).

PR #524, opened 2024-08-28
Abstract rollup addresses into its own package

PJColombo:
### Checklist

### Description
It builds on top of #523.

Moves the rollup addresses into their own package (`@blobscan/rollup`) so it can be consumed by the `web`  and `api` package

The rollup addresses are now required in the web app to implement the rollup filters


xFJA (changes requested):
Same issue than target branch PR https://github.com/Blobscan/blobscan/pull/523#pullrequestreview-2266053770

packages/rollups/package.json
```diff
@@ -0,0 +1,24 @@
+{
+  "name": "@blobscan/rollups",
+  "private": true,
+  "version": "0.2.1",
```
PabloCastellano, line 4: shouldn't it be 0.0.1 or 0.1.0?

packages/rollups/src/index.ts
```diff
@@ -0,0 +1,101 @@
...
+}
+
+export function getAddressByRollup(
+  rollup: Rollup,
+  chainId: number
+): string | null {
+  const rollupMapping = ROLLUP_TO_ADDRESS_MAPPINGS.get(chainId);
+
+  if (!rollupMapping) {
+    return null;
```
PabloCastellano, line 97: this case should be tested

packages/rollups/src/index.ts
```diff
@@ -0,0 +1,101 @@
...
+
+export function getRollupByAddress(
+  address: string,
+  chainId: number
+): Rollup | null {
+  const normalizedAddress = address.toLowerCase();
+  const addressToRollupMapping = ADDRESS_TO_ROLLUP_MAPPINGS.get(chainId);
+
+  if (!addressToRollupMapping) {
+    return null;
```
PabloCastellano, line 84: this case should be tested

packages/rollups/src/index.ts
```diff
@@ -0,0 +1,101 @@
...
+      ),
+    ]
+  )
+);
+
+export function getChainRollups(chainId: number): Rollup[] {
+  const addressToRollupMapping = ADDRESS_TO_ROLLUP_MAPPINGS.get(chainId);
+
+  if (!addressToRollupMapping) {
+    return [];
```
PabloCastellano, line 70: this case should be tested

packages/rollups/src/index.ts
```diff
@@ -0,0 +1,101 @@
+import { Rollup } from "@blobscan/db/prisma/enums";
+
+export const ADDRESS_TO_ROLLUP_MAPPINGS: Map<
+  number,
+  Map<string, Rollup>
+> = new Map([
+  // Mainnet
+  [
+    1,
+    new Map([
```
PabloCastellano, line 10: Why are you using Map() now?

PabloCastellano (changes requested):
Left some questions. Also the packages list in the [documentation](https://docs.blobscan.com/docs/codebase-overview) should be updated

packages/rollups/src/index.ts
PJColombo: It offers the `entries()` function, which allows you to retrieve both the key and value simultaneously. This makes building inverse mappings from rollup to address more straightforward.

packages/rollups/package.json
PJColombo: Yes

PR #525, opened 2024-08-28
feat: Create `Slot` filter

### LINKED ISSUE #434: Add filters to blobs, blocks and transactions pages

### LINKED ISSUE #468: Create a slots filter
Create a slots filter that allows you to filter blocks by a slot range.

The component should be comprised of two numeric inputs joined by a vertical bar.

The text fields should display the following hints: "Start slot", "End slot"

xFJA:
### Checklist

### Description

- Add a `slot` range input to be able to filter in the `blocks` view table by start and end slot number
- Add `placeholder` props to `RangeInput` to be reused in `slot` and `block number` filters

### Related Issue 
- Closes #468 
- Closes #434 

### Screenshots

image


apps/web/src/pages/blocks.tsx
```diff
@@ -218,7 +218,7 @@ const Blocks: NextPage = function () {
       paginationData={{ pageSize: ps, page: p }}
       isExpandable
       rowSkeletonHeight={44}
-      tableTopSlot={<Filters />}
+      tableTopSlot={<Filters enableSlotFilter />}
```
PabloCastellano, line 221: Is this only enabled for blocks on purpose?

apps/web/src/components/RangeInput.tsx
```diff
@@ -5,12 +5,16 @@ export type NumberRange = [bigint | undefined, bigint | undefined];
 export type RangeInputProps = {
   value: NumberRange;
   error?: string;
+  startPlaceholder?: string;
+  endPlaceholder?: string;
```
PabloCastellano, line 9: they should be mandatory, right?

xFJA: The `placeholder` prop is usually optional in any input. We can have scenarios where we want to use a `label` instead of a `placeholder`

apps/web/src/pages/blocks.tsx
xFJA: Yes! You can check the ticket info :+1:

apps/web/src/utils/validators.ts
```diff
@@ -0,0 +1,31 @@
+import type { NumberRange } from "~/components/RangeInput";
+
+export const validateRange = (
```
xFJA, line 3: are we using this func? I think this was part of previous implementation. Should we delete it?

xFJA:
I found quite strange the right alignment on the left side of the range inputs. Could we keep the default left alignment aligned with the rest of the inputs?

image

apps/web/src/utils/validators.ts
PJColombo: Oops, this is a leftover file from the main branch merge.

luis-herasme (approved):

PabloCastellano (changes requested):
It is weird I can't filter only blobs in slot 9727047 like this. If I enter 9727047-9727048 it's fine.

image

This might be an unrelated issue for this PR, but look at the pagination menu weirdness. UPDATE: Opened issue https://github.com/Blobscan/blobscan/issues/535

xFJA:
> It is weird I can't filter only blobs in slot 9727047 like this. If I enter 9727047-9727048 it's fine.
> 
> image
> 
> This might be an unrelated issue for this PR, but look at the pagination menu weirdness.

This error is related to the current filter implementation, similar to the issue related to pagination (not calculating the total items so we have weird behaviours on the table pagination too). I think we could create a ticket to don't forget about this issue and take it further

xFJA:
The interval used in the `slots` filter has been changed to fully closed so now, if you enter in the range the same number, it should show only 1 item with the slot number equal to the number in the range 👍 

The issue related to the total items will be handled in the ticket created 👍

PabloCastellano (approved):
LGTM

PR #537, opened 2024-09-04
Fix Starknet's blob decoding

### LINKED ISSUE #516: Starknet blob decoding contains wrong data
https://blobscan.com/blob/0x0150c4fc905316b21d5c6666866c78303d2c803375f05fce4e5b86698c695a72

image

We made a first implementation using [majin-blob](https://github.com/AbdelStark/majin-blob) that is already running in production. However some data look fishy:

* Contract address doesn't look like it
* Nonce doesn't look correct
* Some more data can probably be decoded
* Storage updates button is almost invisible
* Storage updates are also be wrong (keys should be incremental numbers instead of random hex numbers)
* 'Original Data' switcher should better say 'Raw data'

PJColombo:
### Checklist

### Description
Resolves #516

Some data was not correctly displayed.

Note: this PR does not add support for Starknet [Bolt upgrade](https://www.starknet.io/blog/bolt-version-upgrade/) (since August 28th).

### Screenshots
newDecoding


apps/web/src/utils/number.ts
```diff
@@ -98,3 +98,7 @@ export function cumulativeSum(arr: number[] | bigint[]): number[] | bigint[] {
     );
   }
 }
+
+export function hexToBigInt(hex: string): string {
```
xFJA, line 102: Should we add UTs here?

packages/blob-decoder/src/utils.ts
```diff
@@ -3,3 +3,7 @@ export function bigIntToHex(value: bigint | string): string {
 
   return `0x${value_.toString(16)}`;
 }
+
+export function stripHexPrefix(blobData: string) {
```
xFJA, line 7: Should we add UTs here?

wasms/majin-blob-wasm/majin_blob_wasm_bg.js
```diff
@@ -71,13 +71,13 @@ function passStringToWasm0(arg, malloc, realloc) {
...
 }
 
-let cachedInt32Memory0 = null;
+let cachedDataViewMemory0 = null;
 
-function getInt32Memory0() {
-    if (cachedInt32Memory0 === null || cachedInt32Memory0.byteLength === 0) {
-        cachedInt32Memory0 = new Int32Array(wasm.memory.buffer);
+function getDataViewMemory0() {
+    if (cachedDataViewMemory0 === null || cachedDataViewMemory0.buffer.detached === true || (cachedDataViewMemory0.buffer.detached === undefined && cachedDataViewMemory0.buffer !== wasm.memory.buffer)) {
```
xFJA, line 77: I would create const with meaningful names to improve code readability in this method

xFJA:
_[picky]_

I've noticed the header styles for the table have changed and the `light` theme header background looks violet now. I think transparent looked ok and avoid too many colors when you open an inner table like the `txs` views

Anyway, I think we could give to the first and last header cell rounded corners to be aligned with rest of components

image

xFJA:
Related to my previous comment regarding the [header table ](https://github.com/Blobscan/blobscan/pull/537#issuecomment-2330991696), I think we should back to the previous background because now the headers look weird, without any visible border. The texts seem to be floating in the containers


image

xFJA:
I've checked the responsiveness in the `blobs` view after the changes and found some weird behavior (too much height for the header) in small screens:
image

apps/web/src/utils/number.ts
PabloCastellano: I don't think we need them

PabloCastellano (commented):
Can you add the screenshots?

PabloCastellano (approved):
LGTM. Can you add the screenshots?

packages/blob-decoder/src/utils.ts
PJColombo: No need

wasms/majin-blob-wasm/majin_blob_wasm_bg.js
PJColombo: There's no need. These are the resulting JavaScript wrapper files generated during the wasm build process.

PabloCastellano:
Change Original Data -> Raw Data

PR #544, opened 2024-09-05
All stats

### LINKED ISSUE #432: Unify all stats and metrics pages into one single page
DRAFT:

image

- All metrics cards need to be on top
- All charts need to be below 
- Add  a dropdowns that allows the user to filter charts by type which can be one of the following (`All`, `Blob`, `Block` and `Transaction`). It defaults to `All`

Ref: https://base.blockscout.com/stats

luis-herasme:
### Checklist

### Description
Added a page with all the stats from the blobs, blocks and transactions.

The spacing between the charts was also updated to improve consistency.

### Related Issue
Closes #432

### Screenshots
image

Spacing:
Before:
image

After:
image


xFJA (approved):
I would give more space between the `blobs`,`blocks` and `transactions` chart sections as there are many elements in the view

apps/web/src/pages/stats/all.tsx
```diff
@@ -0,0 +1,21 @@
...
+const AllStats = function () {
+  return (
+    <div className="flex flex-col gap-8">
+      <div className="flex flex-col gap-4">
+        <BlobStats />
+      </div>
+      <div className="flex flex-col gap-4">
+        <BlockStats />
+      </div>
+      <div className="flex flex-col gap-4">
```
xFJA, line 14: Should we create a const for the common `classNames`?

xFJA (commented):
GJ 💪

PJColombo (changes requested):
I’ve updated issue #432 with a draft of the design I had in mind for this page.

Also, instead of fetching the stats individually by entity can we use [`getAllOverallStats`](https://github.com/Blobscan/blobscan/blob/main/packages/api/src/routers/stats/getAllOverallStats.ts) to retrieve the metrics and create a similar one called `getAllDailyStats` to retrieve the chart data?

PabloCastellano (commented):
Not sure if this is the final version but I would only leave the Stats link, no dropdown on click, and link to All metrics .

image

PJColombo:
> Not sure if this is the final version but I would only leave the Stats link, no dropdown on click, and link to All metrics .
> 
> image

Yes, I'd call it "Stats" and not redirect to `/stats/all`

PabloCastellano (commented):
Pending changes:

- Transactions label does not fit into dropdown
- More space between sections
- Remove Stats from section headers (Transaction Stats -> Transactions)
- Dropdown: pluralize names
- Do not round values metrics. Numbers are too exact.
- Set dropdown to '6 months' by default 
- Move filters into its own Card

PabloCastellano (approved):

PabloCastellano (approved):
Pre-approved. Just fix this small thing:

> Remove Stats from section headers (Transaction Stats -> Transactions)

Pluralize also the section headers

Issue #535, opened 2024-09-04
Pagination component is not reset after filtering


PabloCastellano:
From [this comment](https://github.com/Blobscan/blobscan/pull/525#pullrequestreview-2280155867).

My filter returns 0 results but the Pagination component shows lots of pages of results.

![image](https://github.com/user-attachments/assets/772876e5-058a-4fb0-ba1b-53292b3de102)

Issue #528, opened 2024-08-30
Avoid aggregating the total items associated with a rollup when fetching elements filtered by that rollup


PJColombo:
When fetching multiple blobs or transactions filtered by a rollup, an additional Prisma [call](https://github.com/Blobscan/blobscan/blob/a8b94e2dc819d54ff2752960cb87d9ba96399c81/packages/api/src/routers/blob/getAll.ts#L80) is made to count all rows associated with that rollup. This approach is highly inefficient, as it occurs with every request, significantly impacting performance.

We should consider updating this logic to retrieve the total number of elements associated with the rollup from the stats table, once rollup stats support is implemented.

PR #567, opened 2024-09-12
fix(web): fixed blob pagination count

luis-herasme:
### Checklist

### Description
Fixed blob pagination count

### Related Issue
Closes [#535](https://github.com/Blobscan/blobscan/issues/535)
[#528](https://github.com/Blobscan/blobscan/issues/528)

### Screenshots

Before:
image

After:
image

### DIFF
--- a/.changeset/selfish-melons-guess.md
+++ b/.changeset/selfish-melons-guess.md
@@ -1,0 +1,5 @@
+---
+"@blobscan/api": patch
+---
+
+Fixed pagination count for blob search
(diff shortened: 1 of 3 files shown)


PabloCastellano (approved):

xFJA (commented):
I have found some scenarios where the pagination count is not working:

**Scenery 1**
1. Go to Txs page
2. Select a date range where there is no records and filter

The pagination is not right and shouldn't allow navigate

image

**Scenery 2**
1. Go to Blocks page
2. Select a slot range with the same value (the result should be 1 item)

The pagination count is not right

image

PJColombo (approved):

PR #576, opened 2024-09-19
feat: add category and rollup stats

### LINKED ISSUE #518: Calculate stats by category and rollup
Currently our stats page don't take into account what rollup do blobs belong to. So the stats page shows aggregated metrics about the whole chain.

Ideally we want to calculate the stats aggregated for the whole chain but also per rollup/category. This will allow us to display stats per rollup and other use cases such as blobscriptions.

PJColombo:
### Checklist

### Description
Resolves #518

This PR introduces changes to aggregate statistics by category and rollups, and stores them in the daily and overall stats tables. To achieve this, it adds optional `category` and `rollup` columns to these tables, allowing for grouped statistics.

If either of these columns is `null`, it indicates that the row reflects total stats for that grouping. Specifically, when rollup is `null`, the row holds the stats for all rollups. When both category and rollup are `null`, the row represents the overall total stats.

The PR also leverages Prisma’s new [TypedSQL](https://www.prisma.io/typedsql) feature to implement type-safe raw SQL queries.

Additionally, it updates the unit tests to verify the validity and correctness of each aggregated value by comparing them against the aggregation of the fixture data, rather than checking against a snapshot, as we were doing so far.


packages/db/prisma/extensions/base.ts
```diff
@@ -88,33 +47,54 @@ export const baseExtension = Prisma.defineExtension((prisma) =>
...
-                firstBlockNumberAsSender,
+                Prisma.sql`${category.toLowerCase()}::category`,
                 firstBlockNumberAsReceiver,
-                NOW_SQL,
-                NOW_SQL,
+                firstBlockNumberAsSender,
               ]
             )
-            .map((rowColumns) => Prisma.sql`(${Prisma.join(rowColumns)})`);
+            .map((rowColumns) => Prisma.join(rowColumns, ",", "(", ")"));
```
PabloCastellano, line 86: Why that join?

packages/db/prisma/schema.prisma
```diff
@@ -74,15 +74,26 @@ model BlobStoragesState {
   @@map("blob_storages_state")
 }
 
-model Address {
-  address                    String   @id
-  firstBlockNumberAsSender   Int?     @map("first_block_number_as_sender")
+model AddressHistory {
+  address                    String
+  category                   Category @default(OTHER)
```
PabloCastellano, line 79: Category is defined as nullable in other models so I think it would be more coherent to do it here too

packages/syncers/test/__snapshots__/DailyStatsSyncer.test.ts.snap
```diff
@@ -0,0 +1,211 @@
...
+  ],
+  [
+    "2022-10-16",
+    null,
+    null,
+  ],
+  [
+    "2023-05-10",
+    "ROLLUP",
+    null,
```
PabloCastellano, line 138: category is "ROLLUP" but rollup is null

clis/stats-aggregation-cli/test/commands/overall.test.ts
```diff
@@ -27,67 +48,22 @@ describe("Overall command", () => {
   runHelpArgTests(overall, overallCommandUsage);
 
   describe("when incrementing overall stats", () => {
-    it("should increment them until latest indexed block when target block 'latest' is provided", async () => {
+    it.only("should increment them until latest indexed block when target block 'latest' is provided", async () => {
```
PabloCastellano, line 51: debug leftover?

packages/test/src/fixtures/index.ts
```diff
@@ -67,6 +128,98 @@ export const fixtures = {
...
+        );
+        const toHistory = fixtures.addressesHistory.find(
+          (a) => a.address === tx.toId && a.category === tx.category
+        );
+
+        if (!block)
+          throw new Error(`Block with hash "${tx.blockHash}" not found`);
+        if (!fromHistory)
+          throw new Error(
+            `From Address history with id "${tx.fromId}-${tx.category}"  not found`
```
PabloCastellano, line 177:
```suggestion
            `From Address history with id "${tx.fromId}-${tx.category}" not found`
```

packages/db/prisma/schema.prisma
```diff
@@ -74,15 +74,26 @@ model BlobStoragesState {
   @@map("blob_storages_state")
 }
 
-model Address {
-  address                    String   @id
-  firstBlockNumberAsSender   Int?     @map("first_block_number_as_sender")
+model AddressHistory {
```
PabloCastellano, line 77: I don't like the name tbh. How about naming it AddressData, AddressCategoryInfo or AddressByCategoryInfo? :thinking:

packages/db/prisma/migrations/20240916233526_backfill_address_history_table/migration.sql
```diff
@@ -0,0 +1,48 @@
+-- Mgirate data from the address table to the address_history table
```
PabloCastellano, line 1:
```suggestion
-- Migrate data from the address table to the address_history table
```

PabloCastellano (commented):
Good job :muscle: . I've left some comments

clis/stats-aggregation-cli/test/commands/__snapshots__/daily.test.ts.snap
```diff
@@ -3,41 +3,189 @@
...
+    "day": 2023-08-20T00:00:00.000Z,
+    "rollup": null,
+    "totalBlobSize": 2400n,
+    "totalBlobs": 2,
+    "totalUniqueBlobs": 1,
+  },
+  {
+    "category": "ROLLUP",
     "day": 2023-08-20T00:00:00.000Z,
+    "rollup": null,
```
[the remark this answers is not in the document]
PJColombo: > Rollup field shouldn't be null when category is "ROLLUP", right?

For rows in the stats table, this can occur because the row represents an aggregation that includes all rollups, not just a single one. That’s why the rollup value is null.

The same applies to rows where both category and rollup are null; these rows represent total aggregations that include all elements.

clis/stats-aggregation-cli/test/commands/__snapshots__/daily.test.ts.snap
```diff
@@ -3,41 +3,189 @@
...
+  {
+    "category": null,
+    "day": 2023-08-20T00:00:00.000Z,
+    "rollup": null,
     "totalBlobSize": 5900n,
     "totalBlobs": 5,
     "totalUniqueBlobs": 4,
   },
   {
+    "category": null,
```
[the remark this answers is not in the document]
PJColombo: > category should be ROLLUP in this case

Above explanation applies here as well

clis/stats-aggregation-cli/test/commands/__snapshots__/daily.test.ts.snap
```diff
@@ -3,41 +3,189 @@
...
+    "day": 2023-08-20T00:00:00.000Z,
+    "rollup": "ARBITRUM",
+    "totalBlobSize": 1100n,
+    "totalBlobs": 1,
+    "totalUniqueBlobs": 1,
+  },
+  {
+    "category": null,
```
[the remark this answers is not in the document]
PJColombo: > Same

Above explanation applies here as well

clis/stats-aggregation-cli/test/commands/overall.test.ts
PJColombo: yep

packages/syncers/test/__snapshots__/DailyStatsSyncer.test.ts.snap
PJColombo: See [here](https://github.com/Blobscan/blobscan/pull/576#discussion_r1775177526)

packages/db/prisma/schema.prisma
PJColombo: I like `AddressCategoryInfo`

packages/db/prisma/extensions/base.ts
PJColombo: It's a utility function that allows you to join a set of raw sql fragments using a separator and returning a `Prisma.sql` type. This  is the only type allowed by Prisma when writing raw sql queries using `executeRaw()` or `queryRaw()` prisma functions

PR #590, opened 2024-10-09
perf(api): optimise count retrieval

PJColombo:
### Checklist

### Description
It optimizes the retrieval of element counts when filters are applied. By obtaining the total number of elements from the stats tables—specifically when filtering by rollup, category, or timestamp—we can utilize data from the overall or daily stats tables.

### DIFF
--- a/packages/api/src/routers/blob/getCount.ts
+++ b/packages/api/src/routers/blob/getCount.ts
@@ -3,11 +3,11 @@ import { z } from "@blobscan/zod";
 
 import type { Filters } from "../../middlewares/withFilters";
 import {
-  hasCustomFilters,
   withAllFiltersSchema,
   withFilters,
 } from "../../middlewares/withFilters";
 import { publicProcedure } from "../../procedures";
+import { buildCountStatsFilters, hasToPerformCount } from "../../utils/count";
 
 const inputSchema = withAllFiltersSchema;
 
@@ -19,20 +19,9 @@ export async function countBlobs(
   prisma: BlobscanPrismaClient,
   filters: Filters
 ) {
-  if (!hasCustomFilters(filters)) {
-    const overallStats = await prisma.blobOverallStats.findFirst({
-      select: {
-        totalBlobs: true,
-      },
-      where: {
-        AND: [{ category: null }, { rollup: null }],
-      },
-    });
-
-    return overallStats?.totalBlobs ?? 0;
-  }
-
   const {
+    blockNumber,
+    blockTimestamp,
     transactionAddresses,
     transactionCategory,
     transactionRollup,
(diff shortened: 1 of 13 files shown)


packages/api/src/utils/count.ts
```diff
@@ -0,0 +1,50 @@
...
+  const addressFiltersCount = transactionAddresses?.length ?? 0;
+  const hasRollupAsAddress = !!getRollupFromAddressFilter(transactionAddresses);
+  const reorgedFilterEnabled = !!blockType?.some;
+
+  return (
+    blockNumber ||
+    blockSlot ||
+    reorgedFilterEnabled ||
+    addressFiltersCount > 1 ||
+    (addressFiltersCount === 1 && !hasRollupAsAddress)
```
xFJA, line 21: Maybe we could create const for readability

xFJA (approved):
Nice work!! 💯

packages/api/src/routers/blob/getCount.ts
```diff
@@ -19,52 +19,74 @@ export async function countBlobs(
...
+            }
+          : undefined,
+      },
+    });
+  }
+
+  const { fromDay, toDay, category, rollup } = buildCountStatsFilters(filters);
 
-  return prisma.blobsOnTransactions.count({
+  if (fromDay || toDay) {
```
PabloCastellano, line 57: A comment like this would be helpful

```suggestion
  // Counting elements in postgres can become very slow when there are millions of them in a table, so we take advantage of the calculated data used in charts
  if (fromDay || toDay) {
  ```

File: packages/db/prisma/extensions/stats.ts [hunk 3]
```diff
@@ -32,7 +32,9 @@ export const statsExtension = Prisma.defineExtension((prisma) =>
           });
         },
         populate(dailyDatePeriod?: DatePeriodLike) {
-          const { from, to } = toDailyDatePeriod(dailyDatePeriod);
+          const { from: from = MIN_DATE, to: to = new Date() } = dailyDatePeriod
```
PabloCastellano, line 35: IMO it was clearer as it was before, when the function did the checks and set the defaults

packages/api/test/test-suites/filters.ts
```diff
@@ -124,12 +124,6 @@ export function runFilterTests(
       });
     });
 
-    it("should return the correct results when filtering by a 'null' rollup", async () => {
```
PabloCastellano, line 127: why did you remove this test case?

packages/api/test/tx.test.ts
```diff
@@ -94,15 +96,18 @@ describe("Transaction router", async () => {
...
+      await ctx.prisma.transactionDailyStats.populate();
 
       const { totalTransactions } = await caller.tx.getCount({});
 
       expect(totalTransactions).toBe(fixtures.canonicalTxs.length);
     });
 
     runFilterTests(async (filters) => {
+      await ctx.prisma.transactionOverallStats.populate();
+      await ctx.prisma.transactionDailyStats.populate();
```
PabloCastellano, line 110: should we move these two lines to beforeAll()?

packages/api/src/routers/blob/getCount.ts
```diff
@@ -19,52 +19,74 @@ export async function countBlobs(
   prisma: BlobscanPrismaClient,
```
PabloCastellano, line 19: Since this function is now more complex and we use the precalculated values to improve the performance... can you add a comment to this function explaining that? 

Also adding some test cases for this function would be helpful to check that the returned values didn't change.

packages/api/src/routers/tx/getCount.ts
```diff
@@ -16,19 +16,6 @@ const outputSchema = z.object({
 });
 
 export async function countTxs(prisma: BlobscanPrismaClient, filters: Filters) {
-  if (!hasCustomFilters(filters)) {
```
PabloCastellano, line 19: Same as with countBlobs:

- add comment explaining the logic we use now
- add test cases

packages/api/src/utils/count.ts
```diff
@@ -0,0 +1,50 @@
+import { toDailyDatePeriod } from "@blobscan/dayjs";
+
+import { getRollupFromAddressFilter } from "../middlewares/withFilters";
+import type { Filters } from "../middlewares/withFilters";
+
+export function hasToPerformCount({
```
PabloCastellano, line 6:
```suggestion
/* This function determines if the result can be obtained using pre-calculated values from charts or not. */
export function hasToPerformCount({
```

packages/dayjs/src/date.ts
```diff
@@ -51,17 +51,22 @@ export function toDailyDate(
...
 
-  return {
-    from: from ? toDailyDate(from, "startOf").toDate() : MIN_DATE,
-    to: to ? toDailyDate(to, "endOf").toDate() : new Date(),
-  };
+  const datePeriod: DatePeriod = {};
+
+  if (from) {
+    datePeriod.from = toDailyDate(from, "startOf").toDate();
+  }
```
PabloCastellano, line 65:
```suggestion
  } else {
    datePeriod.from = MIN_DATE;
  }
````

File: packages/dayjs/src/date.ts [hunk 10]
```diff
@@ -51,17 +51,22 @@ export function toDailyDate(
...
+
+  if (to) {
+    datePeriod.to = toDailyDate(to, "endOf").toDate();
+  }
```
PabloCastellano, line 69:
```suggestion
  } else {
    datePeriod.to = new Date();
  }
```

packages/db/prisma/extensions/stats.ts
PJColombo: I encountered some errors while running tests, and it took me a while to trace them back to this function setting default date values.

To prevent similar unexpected side effects in the future, I’d prefer to align this function with the`toDailyDate` logic, focusing solely on date normalization without setting default values.

packages/api/src/routers/blob/getCount.ts
PJColombo: Done

packages/api/test/tx.test.ts
PJColombo: These were removed due to latest tests updates

packages/api/test/test-suites/filters.ts
PJColombo: I removed it while testing and forgot to re-add it. I’ve just pushed a commit to add it back.

PR #592, opened 2024-10-12
feat(web): add multiple options to RollupFilter

### LINKED ISSUE #545: Allow to select multiple rollups when filtering elements
We can improve the current filter by using a [combobox](https://headlessui.com/react/combobox)  to allows us to set several rollups

**Task Overview:**

1. **Modify the `withFilter` Middleware**:  
   Update the `withFilter` middleware (found at [withFilters.ts](https://github.com/Blobscan/blobscan/blob/main/packages/api/src/middlewares/withFilters.ts)) by renaming the `rollup` filter to `rollups`. This updated filter should accept a comma-separated string of values. Leverage the existing Zod validation to parse the string into an array of `rollups` enums.

2. **Update Prisma Queries**:  
   Adjust all Prisma queries in the procedures that use these filters (mainly all the `getAll` procedures) to properly handle the new `rollups` filter.

3. **Update Views**:  
   Modify all views that rely on these procedures to send an array of `rollups` instead of a single value.

4. **Create a New Combobox**:  
   Replace the current rollup filter in the UI with a new combobox that supports multiple selection.

5. **Update the Filters Component**:  
   Add the newly created combobox to the `Filters` component to reflect these changes.

xFJA:
### Checklist

### Description
**THIS PR BUILDS ON TOP OF #590**

The `RollupFilter` is able to select multiple rollups. Some changes have been required:

- The `Dropdown` component has a new prop `multiple` that enables the multiple selection
- The `from` query param passed to the `withFilters` in the `API` has been refactored to a string joined by `,` of all from addresses associated to the rollups selected
- Created `useOverflow` hook to detect the overflow in the `RollupFilter` by applying an opacity linear gradient at the right end 
- Add new optional prop to `Option` type in `Dropdown` to be able to customize the component used for the option displayed in the input

### Related Issue 
Closes #545 

### Screenshots 

https://github.com/user-attachments/assets/faeeb8a2-b4fe-411f-bb4e-8dd92540677e


apps/web/src/components/Dropdown/index.tsx
```diff
@@ -6,57 +7,91 @@ import {
...
       <div className="relative">
         <ListboxButton
           className={`relative h-9 ${
             width ?? DEFAULT_WIDTH
           } flex cursor-pointer items-center justify-between rounded-lg border border-transparent bg-controlBackground-light pl-2 pr-8 text-left text-sm shadow-md hover:border hover:border-controlBorderHighlight-light focus:outline-none focus-visible:border-indigo-500 focus-visible:ring-2 focus-visible:ring-white active:border-controlBorderHighlight-dark ui-open:border-controlActive-light dark:bg-controlBackground-dark dark:hover:border-controlBorderHighlight-dark dark:ui-open:border-controlActive-dark`}
         >
-          <div className="truncate align-middle">
-            {selected ? (
-              selected.label ?? selected.value
+          <div
```
PJColombo, line 57: Can we add horizontal scrolling to allow users to navigate through all selected options? Currently, when multiple options are selected, only the ones that fit within the visible area are displayed, while the overflowed options are hidden.

apps/web/src/components/Dropdown/index.tsx
```diff
@@ -6,57 +7,91 @@ import {
...
+import cn from "classnames";
 
+import useOverflow from "~/hooks/useOverflow";
 import { Option } from "./Option";
 
 export interface Option {
   value: string | number;
   label?: ReactNode;
   prefix?: ReactNode;
+  inputDisplay?: ReactNode;
```
PJColombo, line 19: This is not an input so this name is a little bit confusing and not that descriptive. Let's rename it to something more consistent to the rest of the props. Maybe something like `selectedLabel`?

apps/web/src/components/Layouts/PaginatedListLayout.tsx
```diff
@@ -126,6 +126,7 @@ export const PaginatedListLayout: FC<PaginatedListLayoutProps> = function ({
               <div className="flex items-center justify-start gap-2">
                 Displayed items:
                 <Dropdown
+                  multiple={false}
```
PJColombo, line 129: This is not necessary. Prop is `false` by default

packages/api/src/middlewares/withFilters.ts
```diff
@@ -155,7 +155,7 @@ export const withFilters = t.middleware(({ next, input = {} }) => {
       filters.transactionAddresses = [];
       if (from) {
         filters.transactionAddresses.push({
-          fromId: from,
+          fromId: { in: from.split(",") },
```
PJColombo, line 158: Let's add tests to check that this filter works as it suppose to in the following cases:

- when providing multiple values
- when providing empty values (e.g. `",,,,"`, `"value1,,value3"`, etc)

PJColombo, line 158: We should check filter out empty strings when splitting by separator

PJColombo (changes requested):
Nice job.

Left some comments

apps/web/src/components/Dropdown/index.tsx
xFJA: Done 👍

.changeset/old-pianos-help.md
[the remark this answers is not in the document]
xFJA: Done 👍

apps/web/src/components/Dropdown/index.tsx
xFJA: Done 👍

apps/web/src/components/Layouts/PaginatedListLayout.tsx
xFJA: Done! Forgot to remove it after the refactor 👍

packages/api/src/middlewares/withFilters.ts
xFJA: Added 👍

xFJA: Empty strings `from` addresses have been removed 👍 

If we send `,,,,,` as `from` query param, the API should return all the data without this filter instead of an empty list

packages/api/test/test-suites/filters.ts
```diff
@@ -146,12 +162,6 @@ export function runFilterTests(
       });
     });
 
-    it("should return the correct results when filtering by a 'null' rollup", async () => {
```
PabloCastellano, line 149: @xFJA this test still makes sense, right?

packages/api/src/middlewares/withFilters.ts
```diff
@@ -87,6 +89,20 @@ export function hasCustomFilters(filters: Filters) {
   );
 }
 
+export function getRollupFromAddressFilter(
```
PabloCastellano, line 92: This function already exists in packages/api/src/utils/count.ts

Is it a mistake?

PabloCastellano (changes requested):

packages/api/test/test-suites/filters.ts
PJColombo: I think it's a merge-related error

packages/api/src/middlewares/withFilters.ts
PJColombo: Also, a merge-related error

PR #598, opened 2024-10-15
fix(web): add chart fixes

PJColombo:
### Checklist

### Description
It adds the following fixes to the charts on the `/stats` page:

- Display full amounts on the tooltip
- Display amounts without units on the y-axis; moving the unit being used to the chart's title
- Fix an issue where the metric cards was being rendered more than necessary

**BEFORE**
image

image

**AFTER**
image

image


apps/web/src/components/Charts/Transaction/DailyAvgMaxBlobGasFeeChart.tsx
```diff
@@ -44,7 +41,7 @@ export const DailyAvgMaxBlobGasFeeChart: FC<
 
   return (
     <ChartCard
-      title="Daily Avg. Max Blob Gas Fee"
+      title={`Daily Avg. Max Blob Gas Fee (in ${unit.toLowerCase()})`}
```
PabloCastellano, line 44: What cases require it to be lowercase? AFAIK "Gwei" is correct ("gwei" is not).

apps/web/src/components/Charts/Transaction/DailyAvgMaxBlobGasFeeChart.tsx
```diff
@@ -25,13 +25,10 @@ export const DailyAvgMaxBlobGasFeeChart: FC<
     ...buildTimeSeriesOptions({
       dates: days,
       axisFormatters: {
-        yAxisTooltip: (value) => formatWei(value, findBestUnit(value)),
+        yAxisTooltip: (value) => formatWei(value, unit, { displayUnit: true }),
```
PabloCastellano, line 28: Don't we use the best unit anymore?

packages/eth-format/index.ts
```diff
@@ -2,6 +2,9 @@ const ETH_UNITS = { wei: 0, Gwei: 9, ether: 18 };
 
 export type EthAmount = string | number | bigint;
 export type EtherUnit = keyof typeof ETH_UNITS;
+export type FormatOptions = Partial<{
```
luis-herasme, line 5: Now that we sometimes need `formatWei` and `prettyFormatWei` without a unit, I see why this option was added. However, I think a better approach would be to remove the unit display handling from these functions altogether and keep them focused purely on number formatting. Since we already pass the unit to these functions, we can append it directly where needed, like this:

```typescript
`${formatWei(value, unit) ${unit}`
`${prettyFormatWei(value, unit) ${unit}`
```

By keeping the function inputs minimal, we reduce the risk of a combinatorial explosion from too many options, which makes the function easier to test and maintain overall.

apps/web/src/utils/number.ts
```diff
@@ -61,6 +48,34 @@ export function formatNumber(
   );
 }
 
+export function removeCommas(formattedNumber: string): string {
+  return formattedNumber.replace(/,/g, "");
+}
+
+export function parseSuffixedNumber(value: string): [number, string?] {
+  // Remove any leading/trailing whitespace
+  value = removeCommas(value.trim());
```
xFJA, line 57: I would move the `trim()` operation to `removeCommas()` as we are performing the operation again on `L74`. This way, we can pass any value and forget about the whitespace removal

xFJA (approved):

apps/web/src/components/Charts/Transaction/DailyAvgMaxBlobGasFeeChart.tsx
PJColombo: We're still using the best unit. We're just taking it from this hook result a few lines [above](https://github.com/Blobscan/blobscan/blob/2589762d225070b63e3f3639378127a60e243f26/apps/web/src/components/Charts/Transaction/DailyAvgMaxBlobGasFeeChart.tsx#L20) to keep consistency with the label value

packages/eth-format/index.ts
PJColombo: The way I see it: formatting involves transforming a value into a more user-friendly and presentable format, which includes displaying the units.

If you just need to convert an amount, you can use `convertWei`.

To reduce test complexity, we can focus on testing each option separately and ensure the output is correct. For example: when testing the unit display option, we only need to verify that it works correctly, regardless of which unit is being targeted.

PR #608, opened 2024-10-28
Fix display decoded data for Optimism blobs

### LINKED ISSUE #554: Display decoded data for Optimism blobs
PR https://github.com/Blobscan/blobscan/pull/529 adds support for optimism blob decoding for the Ecotone upgrade.

Once this is merged, we need to display this information in the transaction view.

[Example transaction](https://staging.blobscan.com/tx/0x000001bc628a7415eff93e5cb296b0f02a163e24e60453dce163f18113ada54f)

image

TODO:

* [ ] Remove link from `Parent L2 block hash` (the hash is truncated to 20 bytes and currently we don't have any mean to know the full hash and link to it. Maybe also add a trailing '...' to the hash.
* [ ] `L1 origin block hash`: redirect to the block hash. We can find the full hash in our database using the first 20 bytes.
* [ ] Timestamp is incorrect. The timestamp should be subtracted to the block_timestamp.

luis-herasme:
### Checklist

### Description
Fix display decoded data for Optimism blobs

### Related Issue
Closes #554


apps/web/src/pages/tx/[hash].tsx
Comment (PJColombo) [inline apps/web/src/pages/tx/[hash].tsx:1]: This is something I overlooked when reviewing the optimism decoding PR. I believe it would be better to handle optimism decoding on the backend (api package) instead of the client. This approach allows us to serve it on both the web and rest api.

We should implement the decoding in the `getByHash` and `getAll` procedures.

packages/api/src/routers/block/getByPartialHash.ts
PJColombo, line 1: Rather than a procedure this is an utility function. I'd move it to the `utils` folder

packages/api/src/routers/block/getByPartialHash.ts
```diff
@@ -0,0 +1,20 @@
+import { prisma } from "@blobscan/db";
+
+export async function getByPartialHash(hash: string): Promise<string | null> {
```
PJColombo, line 3: Let's rename this to something more descriptive.

```suggestion
export async function getFullBlockHash(partialHash: string): Promise<string | null> {
```

packages/api/src/routers/tx/getByHash.ts
```diff
@@ -65,6 +67,29 @@ export const getByHash = publicProcedure
...
+
+      if (tx.decodedFields) {
+        tx.decodedFields = await addFullBlockHash(tx.decodedFields);
+      }
+
+      return tx;
     }
   );
+
+async function addFullBlockHash(decoded: string): Promise<string> {
```
PJColombo, line 80: Maybe it would make more sense to perform the block hash recovering logic inside of the decoding logic itself.

PJColombo (changes requested):
Well done. I think we can improve it a little bit more.

I left a few request changes.

packages/api/src/routers/block/getByPartialHash.ts
luis-herasme: Done [6fbb32d](https://github.com/Blobscan/blobscan/pull/608/commits/6fbb32d6f80d3546226849ef9fda9484df43d4e1)

packages/api/src/routers/tx/getByHash.ts
luis-herasme: Done [0a276e8](https://github.com/Blobscan/blobscan/pull/608/commits/0a276e85baae58d5b9b55b71a167be9369743415)

apps/web/src/pages/tx/[hash].tsx
Comment (luis-herasme) [reply apps/web/src/pages/tx/[hash].tsx:1]: Done [0a276e8](https://github.com/Blobscan/blobscan/pull/608/commits/0a276e85baae58d5b9b55b71a167be9369743415)

packages/api/src/utils/getFullBlockHash.ts
```diff
@@ -0,0 +1,20 @@
+import { prisma } from "@blobscan/db";
+
+export async function getFullBlockHash(partialHash: string) {
```
PabloCastellano, line 3: I think `autocompleteBlockHash` is more appropriate. What do you think?

apps/web/src/pages/tx/[hash].tsx
```diff
@@ -230,9 +229,9 @@ const Tx: NextPage = () => {
   }
 
   const decodedData =
-    rawTxData && rawTxData.decodedFields
-      ? parseDecodedData(rawTxData.decodedFields)
-      : null;
+    rawTxData?.decodedFields?.type === "optimism"
```
Comment (PabloCastellano) [inline apps/web/src/pages/tx/[hash].tsx:232]: what are the benefits from this change? I liked the previous simplicity because it allowed us to have an agnostic decodedFields json object where we can add more data as we need it without defining any schema.

packages/api/src/blob-parse/optimism.ts
```diff
@@ -28,5 +32,11 @@ export function parseDecodedData(data: string): OptimismDecodedData | null {
     return null;
   }
 
+  const hash = await getFullBlockHash(decoded.data.l1OriginBlockHash);
+
+  if (hash) {
+    decoded.data.l1OriginBlockHash = hash;
+  }
```
PabloCastellano, line 39: It would be nice to log errors here if the full hash is not found.

PabloCastellano:
Tests need to be updated

apps/web/src/pages/tx/[hash].tsx
Comment (PJColombo) [reply apps/web/src/pages/tx/[hash].tsx:232]: This change comes from this [conversation](https://github.com/Blobscan/blobscan/pull/608#discussion_r1819899633)

We already need to parse the `decodedFields` object on the client, and it must comply with a specific type to render the ‘Decoded Fields’ section. Why not handle this directly on the API, so the client isn’t responsible for parsing and the interaction with our API is simplified (via web or our rest api)?

packages/api/src/utils/getFullBlockHash.ts
PJColombo: This comes from this [discussion](https://github.com/Blobscan/blobscan/pull/608#discussion_r1819902336)

I wanted the function name to indicate that we’re fetching the full block hash from an external source. How about `fetchCompleteBlockHash`?

> I think `autocompleteBlockHash` is more appropriate. What do you think?

packages/api/src/blob-parse/optimism.ts
luis-herasme: Done [396b9e9](https://github.com/Blobscan/blobscan/pull/608/commits/396b9e92cc44f824aa877d504469d54debb0b628)

packages/api/src/utils/getFullBlockHash.ts
PabloCastellano: sounds too technical to me

PabloCastellano (approved):
My only nitpick is the name chosen for the function, which sounds too technical, but feel free to merge

packages/api/src/blob-parse/optimism.ts
PabloCastellano: @luis-herasme  should be an error, not a warning

packages/api/src/blob-parse/optimism.ts
```diff
@@ -28,5 +34,15 @@ export function parseDecodedData(data: string): OptimismDecodedData | null {
     return null;
   }
 
+  const hash = await getFullBlockHash(decoded.data.l1OriginBlockHash);
+
+  if (hash) {
+    decoded.data.l1OriginBlockHash = hash;
+  } else {
+    logger.warn(
```
PabloCastellano, line 42:
```suggestion
    logger.error(
```

packages/api/src/blob-parse/optimism.ts
PabloCastellano: I've fixed it now

xFJA (approved):

packages/api/src/utils/autocompleteBlockHash.ts
```diff
@@ -0,0 +1,24 @@
+import { prisma } from "@blobscan/db";
+
+/* Autocomplete a block hash from a truncated version of it.
+   @param partialHash - The first bytes of a block hash.
+   @returns The block hash, if there is a single ocurrence, or null.
+ */
+export async function autocompleteBlockHash(partialHash: string) {
+  const block = await prisma.block.findFirst({
```
PabloCastellano, line 8: @luis-herasme Can you modify it to log an error if there exists more than one occurrence?

PJColombo (approved):
@luis-herasme  Can we add a changeset?

packages/api/src/utils/autocompleteBlockHash.ts
luis-herasme: Done [a83e1bc](https://github.com/Blobscan/blobscan/pull/608/commits/a83e1bca1e2369a5fe987de0088876c869dd2f25)

packages/api/src/utils/autocompleteBlockHash.ts
```diff
@@ -16,9 +17,13 @@ export async function autocompleteBlockHash(partialHash: string) {
...
   });
 
-  if (!block) {
+  if (blocks[0] === undefined) {
     return null;
   }
 
-  return block.hash;
+  if (blocks.length > 1) {
+    logger.warn(`Multiple blocks found for hash ${partialHash}`);
```
PabloCastellano, line 25:
```suggestion
    logger.error(`Multiple blocks found for hash ${partialHash}`);
```

PR #609, opened 2024-10-29
Refactor posthog and fix distinctId

luis-herasme:
### Checklist

### Description
Refactor the `PostHog` integration to follow the official guide better and ensure the application is rendered on both the client and server sides
https://posthog.com/docs/libraries/next-js
image

Also, `PostHog` was not working on the frontend because the `distinctId` was not available.

### DIFF
--- a/apps/web/src/pages/PHProvider.tsx
+++ b/apps/web/src/pages/PHProvider.tsx
@@ -1,0 +1,28 @@
+"use client";
+
+import { useEffect } from "react";
+import posthog from "posthog-js";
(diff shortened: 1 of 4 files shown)


apps/web/src/pages/PHProvider.tsx
PJColombo, line 1: This integration is designed for Nextjs apps that use an app router not a pages one. We should be doing something similar to [this](https://posthog.com/docs/libraries/next-js?tab=Pages+router). 

Also, can we implement the integration in TS?

PJColombo (changes requested):

PR #622, opened 2024-11-08
feat: Add copyable behavior to links/addresses in blob, block and transaction tables

### LINKED ISSUE #621: Add copy icon to blob, blocks and transactions tables addresses
# Issue
The addresses from the `blob`, `block`, and `transactions` tables have been shortened and you need to click on it and check the details to get the whole address value. 

# Approach
Add a copy icon button next to the address to be able to copy it directly from the table.

# Screenshot
We should add the icon right next to the address:
image

xFJA:
### Checklist

### Description

The `Copyable`component has been extended to support links. This way, we can unify all currencies with a copyable behavior into a single component.

The copyable behavior has been added to all links/addresses in the `blob`, `block` and `transactions` tables and its details cards.

#### Motivation
Most of the data is shortened and you need to click and check the details to get the whole value. This way, the user can inteact directly with he table making easier its usage.

### Related Issue 
Closes #621 

### Screenshots
image


PJColombo (changes requested):
Nice job

The `CopyToClipboard` icons in the tables look slightly too large. Could we reduce their size by 1px so they match the height of a line of text?

xFJA:
> Nice job
> 
> The `CopyToClipboard` icons in the tables look slightly too large. Could we reduce their size by 1px so they match the height of a line of text?

I reduced it to `16px` using Tailwind's `-4` space approach rather than the custom `px` value. It also affects the component currencies in the `blob`details, etc

It looks good, aligned with the text:
image

PJColombo (approved):
lgtm 👍

apps/web/src/components/Copyable.tsx
```diff
@@ -0,0 +1,30 @@
...
+
+export interface CopyableProps {
+  href?: string;
+  value: string;
+  tooltipText?: string;
+  children?: React.ReactNode;
+}
+
+export const Copyable: FC<CopyableProps> = ({
+  href,
```
PJColombo, line 15: I'd prefer to remove this prop and keep `Copyable` as generic as possible. If you want to render a link then pass it down but don't as `children`

PJColombo (changes requested):

apps/web/src/components/Copyable.tsx
xFJA: Done!

PJColombo (changes requested):
Links get highlighted when hovering over the copy-to-clipboard icon.

image

I’d prefer for the link to be highlighted only when it’s being hovered over. Let's pass the `Link` as children to the `Copyable` component to prevent this behavior?

PJColombo (approved):
lgtm

Issue #630, opened 2024-11-12
Autocompleting block hash is not working


PabloCastellano:
This new feature used to display decoded data from Optimism Ecotone blobs is not working properly.

Using [this tx](https://blobscan.com/tx/0xffff61c4b20e0fef083407b11f78f0bd7ace25cfb9c0ab898eb75ad76c5f54b6) as example:

![image](https://github.com/user-attachments/assets/ef7b9a86-07ff-442a-8eca-bdd0d12ea7dc)

Clicking on the "L1 origin block hash" link will redirect to an unexisting block: https://etherscan.io/block/0x47403c9454dd9f02d3698e688b3ce1cca7cc4c7b

Additionally:

- [ ] We should not display a link if the hash could not be autocompleted
- [ ] Add optimism example tx data with decoded fields to our fixtures data 
- [ ] If we can successfully autocomplete the L1 origin block, it means we have it in our own db. In that case, we can link to that block on Blobscan instead of Etherscan.
- [ ] Add a link to the “Last L1 origin number” and redirect to Blobscan or Etherscan based on the same condition.

PJColombo:
Resolved in #658

PR #629, opened 2024-11-12
fix: some improvements in autocompleteBlockHash function

PabloCastellano:
### Checklist

- [ ] My change requires a documentation update, and I have done it.
- [ ] I have added tests to cover my changes.
- [x] I have filled out the description and linked the related issues.

PR #631, opened 2024-11-12
feat: update L1 origin block hash handling and add full L1 origin block hash field

luis-herasme:
### Checklist

### Description
If the full block hash cannot be found, this change avoids adding a link to the block hash field, as the URL would be incorrect.

### Related Issue
#630


PJColombo (changes requested):
Left a comment.

Additionally, the [autocompleteBlockHash](https://github.com/Blobscan/blobscan/blob/88055e24a2dbbe151ce69726fc410e9a00a07d9a/packages/api/src/utils/autocompleteBlockHash.ts#L8C67-L9C1) function is not working correctly.

We are filtering hashes that start with the `partialHash` [here](https://github.com/Blobscan/blobscan/blob/88055e24a2dbbe151ce69726fc410e9a00a07d9a/packages/api/src/utils/autocompleteBlockHash.ts#L12). However, this condition will never hold because the partial hash is not prefixed with `0x`, whereas the block hashes are.

PJColombo:
Another request:

Can we move the `autocompleteBlockHash` function back to the [`optimism`](https://github.com/Blobscan/blobscan/blob/f21f6fcb752f6ce5f25339b426e1dcbb29d9d280/packages/api/src/blob-parse/optimism.ts) file?

PJColombo:
A few more changes @luis-herasme  we should do the ui decoded fields based on new info learned about the blobs:

- If we successfully autocomplete the l1 origin block it means we have it on our own db so we can link to that block in blobscan instead of etherscan.
- Let's add a link to the "Last L1 origin number" and redirect to blobscan or etherscan based on previous described condition

apps/web/src/pages/tx/[hash].tsx
```diff
@@ -281,30 +281,28 @@ const Tx: NextPage = () => {
                 },
                 {
                   name: "Parent L2 block hash",
-                  value: (
-                    <div className="flex items-center gap-2">
-                      {"0x" + decodedData.parentL2BlockHash + "..."}
-                    </div>
-                  ),
+                  value: "0x" + decodedData.parentL2BlockHash + "...",
```
Comment (xFJA) [inline apps/web/src/pages/tx/[hash].tsx:284]: I've found `Ox` several times repeated as a raw string. I would create a const for it

xFJA (changes requested):
Missing changeset

apps/web/src/pages/tx/[hash].tsx
Comment (PabloCastellano) [reply apps/web/src/pages/tx/[hash].tsx:284]: Beginning with 0x just means that the string is hexadecimal. I feel like having const for just a two characters string is kind of an overkill

PabloCastellano (approved):

PabloCastellano:
@luis-herasme Can you also apply my changes from https://github.com/Blobscan/blobscan/pull/629 ?

xFJA (approved):

PR #642, opened 2024-11-25
feat(blob-propagation-jobs-cli): add 

PJColombo:
### Checklist

### Description
It fixes an issue where the create command failed to properly iterate over blobs within a specified date range.

Additionally, it introduces support for creating jobs based on a specified block number range.


PabloCastellano (commented):
Another changeset should be created for the following feature:

> Additionally, it introduces support for creating jobs based on a specified block number range.

PR #654, opened 2024-11-28
chore(db): fix seed script

PJColombo:
### Checklist

### Description

It does the following:
- It fixes an issue where the seed script wasn't creating the address category info db entities correctly.
- It also generates blob data for every created blob.
- It improves the script's logs
- It replaces `ts-node` by `tsx` as former resolves modules to commonjs which was giving problems with certain esm-only packages (e.g. `ora`)


xFJA (approved):
Nice job 💯

Issue #650, opened 2024-11-25
Capture usage metrics


PabloCastellano:
* [ ] Add 'network' property to trpc_event (sepolia, holesky)
* [ ] Add 'instance' property to trpc_event (mainnet, staging, sepolia, holesky, gnosis)

Posthog:
* [ ] Configure "URL query parameter converter" https://posthog.com/docs/cdp/url-query
* [ ] Add dashboards breaking down by network and instance
* [ ] Merge https://github.com/Blobscan/blobscan/pull/609

PR #663, opened 2024-12-03
fix: fixes client side PostHog events capturing and adds extra data to the metadata

luis-herasme:
### Checklist

### Description
This update fixes the client-side capturing of `PostHog` events and adds additional data to the metadata.

The issue with client-side `PostHog` event capturing arose because we were attempting to read a server-side environment variable within the `loaded` function.

Additionally, we now save the network name, as well as the `pathname` and `queryString` in the `PostHog` metadata.

### Related Issue (Optional)
#650

### DIFF
--- a/packages/api/src/context.ts
+++ b/packages/api/src/context.ts
@@ -9,11 +9,10 @@ import type {
 import cookie from "cookie";
 import jwt from "jsonwebtoken";
 
-import { BlobPropagator, getBlobPropagator } from "@blobscan/blob-propagator";
-import {
-  BlobStorageManager,
-  getBlobStorageManager,
-} from "@blobscan/blob-storage-manager";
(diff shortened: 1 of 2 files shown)


PJColombo (approved):
I’m not very familiar with PostHog or the integration logic, so I don’t have much to add. If you’ve tested this and confirmed that it’s now capturing client-side events, it looks good to me.

packages/api/src/context.ts
```diff
@@ -119,14 +118,25 @@ export function createTRPCContext(
           );
         }
 
+        let pathname: string | undefined;
+        let query: string | undefined;
+
+        if (opts.req.url) {
```
PabloCastellano, line 124: In what cases is this condition false?

PabloCastellano (approved):
lgtm

PR #679, opened 2025-01-09
fix: include host in $current_url

luis-herasme:
### Checklist

### Description
Include host in $current_url

### Related Issue
#650

### DIFF
--- a/packages/api/src/context.ts
+++ b/packages/api/src/context.ts
@@ -118,12 +118,14 @@ export function createTRPCContext(
           );
         }
 
+        let current_url = opts.req.url;
         let pathname: string | undefined;
         let query: string | undefined;
 
         if (opts.req.url) {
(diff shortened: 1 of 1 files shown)


packages/api/src/context.ts
```diff
@@ -118,12 +118,14 @@ export function createTRPCContext(
           );
         }
 
+        let current_url = opts.req.url;
```
xFJA, line 121:
```suggestion
        let currentUrl = opts.req.url;
```

we should use `camelCase` when using `TypeScript`

xFJA (approved):

packages/api/src/context.ts
luis-herasme: Done [322d1a8](https://github.com/Blobscan/blobscan/pull/679/commits/322d1a871e09dde478b49c09f6b10138cf8bc5f5)

PabloCastellano (approved):

PR #681, opened 2025-01-10
feat: add weavevm support

PJColombo:
### Checklist

### Description
This PR adds support for Weavem by implementing the following changes:
	•	New Blob Storage: Introduced a new blob storage system to retrieve blobs from the WeaveVM API.
	•	UI Enhancement: Added a new UI badge to represent Weavem functionality.


xFJA (commented):
Good job! 💪 💪 

Should we add any seed data to be able to check in local/deployments?

packages/blob-propagator/src/BlobPropagator.ts
```diff
@@ -84,7 +85,18 @@ export class BlobPropagator {
...
+          logger.warn(
+            `Worker processor not defined for storage "${name}"; skipping`
+          );
+        }
+
+        return hasWorkerProcessor;
+      });
     const temporaryBlobStorage = blobStorageManager.getStorage(tmpBlobStorage);
 
     if (!availableStorageNames) {
```
luis-herasme, line 102: `!availableStorageNames` will be false even if the array is empty, we should check the array length:
 
image

packages/blob-propagator/src/BlobPropagator.ts
```diff
@@ -84,7 +85,18 @@ export class BlobPropagator {
     const availableStorageNames = blobStorageManager
       .getAllStorages()
       .map((s) => s.name)
-      .filter((name) => name !== tmpBlobStorage);
+      .filter((name) => name !== tmpBlobStorage)
+      .filter((name) => {
+        const hasWorkerProcessor = !!STORAGE_WORKER_PROCESSORS[name];
```
luis-herasme, line 90: I am not a fan of implicit coercion, but this is a stylistic preference.

We can add an `eslint-rule` to disable it:
https://eslint.org/docs/latest/rules/no-implicit-coercion#boolean

luis-herasme (approved):

packages/blob-propagator/src/BlobPropagator.ts
PJColombo: Well spotted

packages/blob-propagator/src/BlobPropagator.ts
PJColombo: We’ve been using implicit coercion for quite some time now.

I’d prefer to discuss this in a meeting and then create a PR to enable the rule and replace all instances of it, in case we agree to ban it.

packages/blob-propagator/src/BlobPropagator.ts
PJColombo: Added some UTs to check this case as well

PR #695, opened 2025-01-28
Fix stats order

### LINKED ISSUE #687: Stats bar charts are displayed backwards
Image

luis-herasme:
### Checklist

### Description
Stats bar charts are displayed backwards

### Related Issue
Closes  #687

### DIFF
--- a/.changeset/big-bags-warn.md
+++ b/.changeset/big-bags-warn.md
@@ -1,0 +1,5 @@
+---
+"@blobscan/api": patch
+---
+
+Added sorting option to the getDailyStats API endpoint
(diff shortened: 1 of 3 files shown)


PabloCastellano (approved):
:+1:

PJColombo (changes requested):
I think it makes sense to return the stats sorted by the most recent one.

I was thinking we could add a sort query parameter, similar to the one used in endpoints that return multiple elements.

We can use this [sort schema](https://github.com/Blobscan/blobscan/blob/90a05b3e84fe151538411de2ee378cbbaa97a316/packages/api/src/middlewares/withFilters.ts#L84). We can then update the stats page trpc request by adding `sort: "asc"` query param

While we’re at it, we could abstract this functionality into its own middleware,`withSort`.

PJColombo:
Don't forget to add the changeset

PJColombo (approved):
lgtm

PR #697, opened 2025-01-28
refactor(env): replace conditionalRequiredSchema and requiredStorageConfigSchema with requireIfEnvEnabled

luis-herasme:
#### Refactor of Conditional Require Logic for Environment Variables
This PR refactors the logic used to conditionally require environment variables based on other settings. Previously, we used a function called `conditionalRequiredSchema`, which was exclusively invoked by `requiredStorageConfigSchema`. The logic was convoluted and difficult to follow. This refactor simplifies the logic and makes it more intuitive.

#### Changes:
- Replaced conditionalRequiredSchema with a more direct approach using superRefine in Zod schemas.
- Introduced a new utility function requireIfEnvEnabled, which checks if an environment variable is enabled (process.env[<env_name>] === "true") and enforces that another field is required based on this.
- Updated the schema for several environment variables like SWARM_STORAGE_ENABLED and GOOGLE_STORAGE_ENABLED to use this new approach for conditional validation.

### DIFF
--- a/packages/env/index.ts
+++ b/packages/env/index.ts
@@ -26,52 +26,55 @@ function requireIfEnvEnabled(envName: string) {
 export const env = createEnv({
   envOptions: {
     server: {
-      // PostHog
-      POSTHOG_ID: z.string().optional(),
-      POSTHOG_HOST: z.string().default("https://us.i.posthog.com"),
-
-      BEE_ENDPOINT: z
-        .string()
-        .url()
-        .optional()
-        .superRefine(requireIfEnvEnabled("SWARM_STORAGE_ENABLED")),
+      // General
+      BLOBSCAN_API_PORT: z.coerce.number().positive().default(3001),
+      NODE_ENV: nodeEnvSchema.optional(),
+      SECRET_KEY: z.string(),
+      CHAIN_ID: z.coerce.number().positive().default(1),
+      DENCUN_FORK_SLOT: z.coerce.number().optional(),
+      NETWORK_NAME: networkSchema.default("mainnet"),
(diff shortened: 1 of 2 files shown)


PabloCastellano (approved):

packages/env/index.ts
```diff
@@ -12,16 +12,15 @@ import {
...
-  storageName: "FILE_SYSTEM" | "GOOGLE" | "POSTGRES" | "SWARM",
-  schema: T
-) {
-  return conditionalRequiredSchema(
-    schema,
-    process.env[`${storageName}_STORAGE_ENABLED`],
-    "true",
-    `This configuration variable is required when ${storageName} storage is enabled.`
-  );
+function requireIfEnvEnabled(envName: string) {
```
PJColombo, line 15: Let's apply this function to the following google storage env vars: `GOOGLE_STORAGE_BUCKET_NAME`,  `GOOGLE_SERVICE_KEY` and `GOOGLE_STORAGE_API_ENDPOINT`

and to this weavevm env var: `WEAVEVM_STORAGE_API_BASE_URL`

PJColombo (changes requested):
lgtm.

There are some more storage env vars where we can use this new function

packages/env/index.ts
luis-herasme: Done [f9a3a02](https://github.com/Blobscan/blobscan/pull/697/commits/f9a3a022ca50a9c81c186c7075a96ead2593e7b1)

PJColombo (approved):
lgtm

PR #700, opened 2025-01-30
feat(api): add support for fetching blocks by slot

PJColombo:
### Checklist

### Description
It adds a new procedure `getBySlot` which enables consumers to fetch blocks by their slot. It also exposes the procedure as a REST endpoint.

### DIFF
--- a/packages/api/src/routers/block/getByBlockId.ts
+++ b/packages/api/src/routers/block/getByBlockId.ts
@@ -1,73 +1,35 @@
 import { TRPCError } from "@trpc/server";
 
-import type { Prisma } from "@blobscan/db";
 import { hashSchema, z } from "@blobscan/zod";
 
 import {
   createExpandsSchema,
   withExpands,
 } from "../../middlewares/withExpands";
-import type { Filters } from "../../middlewares/withFilters";
 import {
   withFilters,
   withTypeFilterSchema,
 } from "../../middlewares/withFilters";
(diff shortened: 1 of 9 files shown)


packages/api/src/routers/block/getByBlockId.ts
```diff
@@ -19,37 +21,54 @@ import {
...
+  const blockNumberOrSlotRes = blockNumberOrSlotSchema.safeParse(id);
+
+  if (blockNumberOrSlotRes.success) {
+    return {
+      [isSlot ? "slot" : "number"]: blockNumberOrSlotRes.data,
+      transactionForks: filters.blockType,
+    };
+  }
+
+  const blockHashResult = blockHashSchema.safeParse(id);
```
luis-herasme, line 60: The `blockHashSchema` will never succeed if the `blockNumberOrSlotSchema` has already failed.

PabloCastellano (commented):
Instead of expanding the scope of `getByBlockId`, why not create a new method?

luis-herasme:
> Instead of expanding the scope of `getByBlockId`, why not create a new method?

I prefer this too

PJColombo:
> Instead of expanding the scope of `getByBlockId`, why not create a new method?

Because I want to expose this procedure as a REST endpoint, having two endpoints that essentially perform the same function—returning a block based on an ID—feels redundant.

PR #762, opened 2025-04-27
Do not return blob data in api response

### LINKED ISSUE #761: Do not return blob data in api response
As part of our cost savings efforts, we need to make some changes in our (free) API.

Some APIs fetch the data from one of the storages and return it in the response, so we end up paying twice the egress costs (bucket --> vm --> end user). 

* [ ] **GET /blobs/{id}**

Remove the `data` field containing the blob data. This will reduce the response size from ~256KB to 1KB and will be much faster.

* [ ] **GET /blobs/{id}/data**

Add environment variables to enable or disable this endpoint, and optional authentication.

luis-herasme:
### Checklist

### Description
**This PR introduces a BREAKING CHANGE**

It does the following:
- Removes the data field from all API responses related to blobs.
- Adds API key authorization to the getBlobDataByBlobId procedure, configurable via the BLOB_DATA_API_KEY environment variable.
- Adds ability to disable this endpoint and avoid egress charges by setting BLOB_DATA_API_ENABLED=false
- Updates the web app to retrieve blob data from the blob storage reference URLs instead of trpc procedures

These changes are aimed at reducing response payload sizes and lowering infrastructure costs.

### Related Issue

Closes #761


PabloCastellano (approved):
GJ

apps/docs/src/app/docs/environment/page.md
```diff
@@ -12,6 +12,7 @@ nextjs:
...
 | `DIRECT_URL`                           | Direct connection to the database used by Prisma CLI for e.g. migrations.                   | Yes      | (empty)                                                                                                                                                                                                                                                        |
+| `BLOB_DATA_API_KEY`                    | API key used to authenticate requests to retrieve blob data from the blobscan API           | No       | (empty)                                                                                                                                                                                                                                                        |
```
PabloCastellano, line 15: The webapp doesn't need this new variable, does it?

apps/web/src/env.mjs
```diff
@@ -76,6 +77,7 @@ export const env = createEnv({
     DATABASE_URL: process.env.DATABASE_URL,
     DIRECT_URL: process.env.DIRECT_URL,
     FEEDBACK_WEBHOOK_URL: process.env.FEEDBACK_WEBHOOK_URL,
+    BLOB_DATA_API_KEY: process.env.BLOB_DATA_API_KEY,
```
PabloCastellano, line 80: I think an extra variable BLOB_DATA_API_ENABLED would be useful for our case. We are incurring in extra charges because of this endpoint, so we want to disable it, not even keep it behind authentication. However, if another team wants to spin up a blobscan instance, they can decide if enable/disable and authenticated/public.

PabloCastellano (commented):
looks good, I 've left some comments

apps/web/src/env.mjs
PabloCastellano: added

PR #778, opened 2025-05-21
Fix broken Indexer link in API documentation

GarmashAlex:
Corrected the relative link to the Indexer documentation in apps/docs/src/app/docs/api/page.md to ensure proper navigation and resolve the "Cannot find file" error. The link now correctly points to the ../indexer directory.


PabloCastellano:
I could not reproduce the problem. Don't hesitate to open and share steps to reproduce

PR #801, opened 2025-06-06
fix: run SwarmStampSyncer when a batch id has been previously configured

PabloCastellano:
### Checklist

### Description
The Swarm TTL is not updated if SWARM_STORAGE_ENABLED=false, e.g. when Swarm storage is temporarily disabled.

### DIFF
--- a/apps/rest-api-server/src/syncers.ts
+++ b/apps/rest-api-server/src/syncers.ts
@@ -2,6 +2,7 @@ import type { PublicClient } from "viem";
 import { createPublicClient, http } from "viem";
 import * as chains from "viem/chains";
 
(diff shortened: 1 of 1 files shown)


apps/rest-api-server/src/syncers.ts
```diff
@@ -2,6 +2,7 @@ import type { PublicClient } from "viem";
 import { createPublicClient, http } from "viem";
 import * as chains from "viem/chains";
 
+import { prisma } from "@blobscan/db";
```
PabloCastellano, line 5: should be a procedure

PabloCastellano:
We have dropped Swarm support

PR #804, opened 2025-06-15
feat(blob-storage): update blob storage to use binary data and set content type to application/octet-stream

### LINKED ISSUE #753: Upload blobs as binary
We are uploading the ascii representation which takes 2x the size.

luis-herasme:
### Checklist

### Description
Closes  #753


packages/blob-storage-manager/src/storages/SwarmStorage.ts
```diff
@@ -97,11 +97,14 @@ export class SwarmStorage extends BlobStorage {
   }
 
   async #sendToBeeNode(versionedHash: string, data: string) {
+    const buffer = Buffer.from(data);
```
xFJA, line 100: We are doing the same operation in `sendToChunkstorm`. We could maybe create the buffer in `_storeBlob` and pass the binary directly to both send functions

PJColombo (changes requested):
We should update the google storage to upload binary data as well

PR #992, opened 2026-06-23
fix(db): add IPFS case to blob URL computed field

PabloCastellano:
### Checklist

### Description
  
The blobDataStorageReference computed url field (defined in packages/db/prisma/extensions/computed-fields.ts) had cases for every storage backend — GCS, Postgres, Swarm, WeaveVM, S3 — but was missing IPFS. As a result, any blob stored on IPFS would have url: undefined, causing the IPFS storage badge in the /blobs list (and blob detail pages) to render without a clickable link.

After the IPFS storage backend was added (#985), blobs uploaded via blobscan-ipld (POST /blobs/ipfs-references) were correctly stored in blob_data_storage_reference with storage = 'IPFS', but  the IPFS badge never appeared in the UI because the URL was never computed. This fix completes the integration so the badge links directly to the blob on the configured IPFS HTTP gateway (e.g.  https://ipfs.filebase.io/ipfs/<cid>).


PJColombo (approved):
lgtm

PR #994, opened 2026-06-23
fix(storage): retry health check on transient startup failures

PabloCastellano:
### Checklist

### Description

Adds retry-with-backoff to BlobStorage.healthCheck() — the shared base-class method that every storage backend's create() calls before being added to the manager:

- Up to 4 attempts (1 initial + 3 retries)
- Exponential backoff with jitter (~500ms / ~1s / ~2s × 0.5–1.0), matching the existing retry style in IpfsStorage
- Only after all attempts are exhausted does it throw the original "Storage is not reachable" error, preserving the underlying failure as cause

The REST API server aborts startup when its primary blob storage fails its initial reachability check — even when the failure is a transient network hiccup rather than a real misconfiguration. We observed this in staging:

```
  StorageCreationError: GoogleStorage failed: Creation failed: Storage is not reachable
    cause: FetchError: Invalid response body while trying to fetch
           https://www.googleapis.com/oauth2/v4/token: Premature close
  ...
  Primary blob storage "GOOGLE" was not created.
```

The root cause was a premature connection close while fetching a GCS OAuth2 token — a one-off transport blip, not a credentials or config problem (which would surface as a 4xx). Because GOOGLE is the primary storage,
  createBlobPropagator then refused to start the server. Meanwhile other services (e.g. eth-price-syncer) had no trouble reaching the internet, confirming the failure was transient and storage-specific.

A single network blip at boot should not take down the API.

### DIFF
--- a/.changeset/storage-healthcheck-retry.md
+++ b/.changeset/storage-healthcheck-retry.md
@@ -1,0 +1,5 @@
+---
+"@blobscan/blob-storage-manager": patch
+---
+
+Retry blob storage health checks on startup so a transient network failure (e.g. a premature connection close while fetching a GCS auth token) no longer aborts boot. The reachability probe now uses up to 4 attempts with exponential backoff and jitter before declaring a storage unreachable.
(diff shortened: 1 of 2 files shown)


PabloCastellano:
superseeded by 0e9163142cc5286ca48da348bf906c9807f0dd68

PR #995, opened 2026-06-24
fix(rollups): match blob poster addresses case-insensitively

daixihegu:
### Checklist

### Description

Update `RollupRegistry.getRollup` to compare blob poster addresses case-insensitively.

Some registry entries are stored as checksum/mixed-case addresses, while transaction addresses can be lowercased before lookup. This could cause `getRollup` to miss a known blob poster and return `null`. The lookup now normalizes both sides before comparing, without changing the stored registry values.


daixihegu:
@PabloCastellano  @PJColombo Hi, Could you please review this PR at your convenience? Thank you very much.

PR #1007, opened 2026-07-16
fix(web): decode Postgres blob data as hex bytes

pengqima:
### Checklist

### Description

Fix the blob data proxy response for Postgres-backed storage references.

The Postgres blob data endpoint returns a JSON string containing `0x`-prefixed hexadecimal data. The proxy previously passed the hexadecimal characters to `Buffer.from()` without specifying an encoding, causing them to be interpreted as UTF-8 text instead of the original binary bytes.

This change decodes the value using the `hex` encoding, matching the existing handling for text-based storage responses.

For example, `0x0001feff` was previously returned as the bytes representing the ASCII string `0001feff` (`3030303166656666`) instead of the expected binary bytes (`0001feff`).

A regression test now verifies that Postgres-backed responses preserve the original blob bytes.

### DIFF
--- a/.changeset/fix-postgres-blob-data-decoding.md
+++ b/.changeset/fix-postgres-blob-data-decoding.md
@@ -1,0 +1,5 @@
+---
+"@blobscan/web": patch
+---
+
+Fixed Postgres-backed blob data being returned as UTF-8 encoded hexadecimal characters instead of the original binary bytes.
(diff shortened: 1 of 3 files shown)


pengqima:
@PabloCastellano @PJColombo Hi, Could you please review this PR at your convenience? Thank you very much.

PR #1011, opened 2026-08-03
fix(api): correct stats downsampling dimension estimation

grandpig:
### Checklist

### Description

- Fix stats downsampling dimension estimation
- Add regression tests for category and rollup filter combinations

The downsampling heuristic currently estimates dimensions by multiplying category and rollup counts:

```ts
categoriesCount * rollupsCount
```

This does not match the actual query behavior, which uses category and rollup filters as separate alternatives (OR), rather than a Cartesian product.

It also causes category-only and rollup-only requests to estimate zero dimensions, so downsampling may never be triggered for those requests.


grandpig:
@PJColombo Just a gentle follow-up in case this got buried. This is a small fix to the stats downsampling heuristic, with regression coverage for category-only, rollup-only, and combined filters. Would appreciate a look when you have time. Thanks!

grandpig:
@PJColombo  Could you please take a look when you have a chance? Thanks!

PR #1012, opened 2026-08-24
fix(api): end telemetry spans when middleware throws

wangjingshuiku:
### Checklist

- [ ] My change requires a documentation update, and I have done it. (Not applicable: no public API or configuration changes.)
- [ ] I have added tests to cover my changes. (No new test was added; the API package type-check passes.)
- [x] I have filled out the description and linked the related issues. (No related issue.)

### Description

End the active middleware span from a `finally` block so it is closed on both successful requests and exceptions from downstream middleware. Return values and error propagation remain unchanged.

The previous implementation called `span.end()` only after `next()` resolved. If `next()` threw, the function exited before ending the span, leaving incomplete telemetry.

Validation: `pnpm --filter @blobscan/api type-check`

### Related Issue (Optional)

None.

Not applicable.


wangjingshuiku:
@PJColombo  Could you please take a look when you have a chance? Thanks!

